PR #1, opened 2024-05-03
SOF-1866: Setup barebone package structure

KvelaGorrrrnio:
The repository is setup for developing a TypesScript application targeting a node (20+) platform. Highlights for the setup:
- TypeScript compiler (`tsc`) is used for building. The TS configuration is nearly identical to sofie-server configuration. The code compiles to a CommonJS setup, while the tools are setup to run as ES modules.
- `ts-node` and `nodemon` are used for setting up the development server.
- Vitest is used as the test runner and assertion library.
- Eslint is used for linting. There are more formatting rules applied here than in the sofie server and sofie web client.
- `yarn npm audit` is used for dependency validation on moderate threads for production dependencies.
- `license-checker` is used for ensuring that we only use packages with useable licenses.


.gitignore
```diff
@@ -0,0 +1,13 @@
+node_modules/
```
LindvedKrvang, line 1: Would be nice to also add `.idea/` so we ignore JetBrains specific files.
Same for `.vscode`.

package.json
```diff
@@ -0,0 +1,43 @@
+{
+  "name": "sofie-tv2-server",
+  "version": "0.0.0",
+  "repository": "git@github.com:tv2/sofie-tv2-server.git",
+  "author": "Anders Frederik Jørgensen <afjo@tv2.dk>",
```
LindvedKrvang: Do we want authors?

LindvedKrvang (approved):

package.json
RasmusAlbrektsen: A good author, would recommend.

src/name-generator.ts
```diff
@@ -0,0 +1,5 @@
+export class NameGenerator {
+  public getName(): string {
+    return 'Sofie TV 2 server'
```
RasmusAlbrektsen, line 3: Should move this string to a constant, this is kind of sloppy Anders. Seriously dude.

RasmusAlbrektsen (approved):

.gitignore
KvelaGorrrrnio: That should be in your global gitignore, not in the repository :)

package.json
KvelaGorrrrnio: I don't think we want a single author. I have moved it to contributors.

PR #11, opened 2024-08-20
SOF-1869 + SOF-1871: Set up HTTP and WebSocket proxy

KvelaGorrrrnio:
- Adds Fastify server with proxy configuration for HTTP requests and WebSocket connections.
- ESLint configuration is updated to not allow `console` statements, and adds logger with structured output.
- Sample code is removed, which is the reason that `--passWithNoTests` is added to the test scripts in the package.json file.


src/index.ts
```diff
@@ -1,5 +1,16 @@
-import { NameGenerator } from './name-generator'
+import { ProxyServer } from './interfaces/proxy-server'
+import { FastifyServer } from './fastify-server'
+import { ConsoleLogger } from './console-logger'
+import { Logger } from './interfaces/logger'
 
-const nameGenerator: NameGenerator = new NameGenerator()
+const SOFIE_REST_URL: string = 'http://localhost:3005'
```
RasmusAlbrektsen, line 6: *Cough cough*, might we want different names for the variables? Or would we rather come back to that?

RasmusAlbrektsen (approved):

src/index.ts
KvelaGorrrrnio: I would rather that we do it when the repo and package is renamed to avoid an in-between state.

LindvedKrvang (approved):

PR #14, opened 2024-09-24
Sof 2226/create panel controller

LindvedKrvang:
Note: Dependent on https://github.com/tv2/sofie-tv2-server/pull/13

Setup annotations for REST controllers. The setup for annotations is the same as for Alba-server.
Expose the endpoint `/api/panels/physicalPanelLayouts` that fetches the collection of `PhysicalPanelLayout`.

Note: I changed the nodemon debug server from port `9229` to port `9230` so it doesn't clash with the debug port of Alba-server.


src/presentation/index.ts
```diff
@@ -0,0 +1,20 @@
+import { ProxyServer } from './interfaces/proxy-server'
+import { FastifyServer } from './fastify-server'
+import { ConsoleLogger } from '../model/console-logger'
+import { Logger } from '../model/interfaces/logger'
+import { BaseController } from './controllers/base-controller'
+import { ControllerFacade } from './facades/controller-facade'
+
+const SOFIE_REST_URL: string = 'http://localhost:3005'
```
RasmusAlbrektsen, line 8: I know that we talked about this being a task for later at not e.g. in this specific PR. I just want to make a point of the fact that we should really rename to Alba in this repo before we make it a bigger task for ourselves :-)

RasmusAlbrektsen (dismissed):

MadsThy (dismissed):

PR #15, opened 2024-09-26
Sof 2228/create panel layout repository

LindvedKrvang:
Initial setup of our a data-access layer. Accesing the MongoDB is mostly copied from `sofier-server`. Note, that we are accessing a different database, `alba-tv2-server`, within the MongoDB instance.

Added types for a `ConfiguredPanalLayout` and expose a method in the `ConfiguredPanelLayoutRepository` to save it.

Also in this PR:
- Downgraded eslint to 8.56.0 since it was incompatiable with the versions of our plugins (this was the easiest solution...)


RasmusAlbrektsen (approved):

src/data-access/interfaces/uuid-generator.ts
```diff
@@ -0,0 +1,3 @@
+export interface UuidGenerator {
```
KvelaGorrrrnio, line 1: From our chat last week about having abstractions when we need them: Will we ever have multiple UuidGenerator implementations?

src/data-access/mongo/mongo-database.ts
```diff
@@ -0,0 +1,67 @@
...
+    this.db = this.client.db(MONGO_DB_NAME)
+    this.logger.info(`Connected to database: ${this.db.databaseName}`)
+  }
+
+  public getCollection<Model extends MongoId>(collectionName: string): Collection<Model> {
+    this.assertDatabaseConnection(this.db)
+    return this.db.collection(collectionName)
+  }
+
+  private assertDatabaseConnection(database?: Db): asserts database is NonNullable<Db> {
```
KvelaGorrrrnio: Is the NonNullable type wrapper necessary? The type `Db` doesn't seem to be nullable.

src/model/interfaces/configured-input.ts
```diff
@@ -0,0 +1,40 @@
+import { InputDataType, InputType } from '../enums/panel-enums'
+import { HttpMethod } from '../enums/http-method'
+
+export type ConfiguredInput = ConfiguredButtonInput | ConfiguredFaderInput
+
+export interface DefaultConfiguredInput {
+  type: InputType
+  data: InputData
+}
```
KvelaGorrrrnio: I am trying to weigh `ConfiguredInput` vs `InputConfiguration`. My brain is not entirely sure about what _configured input_ mean and doesn't mean. It sounds like it is something preconfigured, whereas this refers to the configuration object for a specific panel input.

I assume that `Default` refers to this being the base from which the other configurations extend. When I read this I assumed this was some kind of fallback data structure that could be used. It can be considered if this should use another word than `Default`, e.g. `Base` as suffix.

This might just be me, but I think that `InputConfigurationBase` and `InputConfiguration` have better semantics.
In continuation of this, having both `Button` and `Input` seems a bit redundant and changes the semantics for me. I will suggest that `ConfiguredButtonInput` is renamed to `ButtonConfiguration` and that `ConfiguredFaderInput` is renamed to `FaderConfiguration`.

src/model/interfaces/configured-panel-layout.ts
```diff
@@ -0,0 +1,10 @@
+import { PanelModel, PanelType } from '../enums/panel-enums'
+import { ConfiguredInput } from './configured-input'
+
+export interface ConfiguredPanelLayout {
```
KvelaGorrrrnio: Same comment as the other one about the `Configured` prefix. Personally I am in favor of having it as a `Configuration` suffix.

KvelaGorrrrnio (approved):

src/data-access/interfaces/uuid-generator.ts
LindvedKrvang: Probably not, but I don't recall that we decided to stop doing Dependency Inversion.

src/data-access/mongo/mongo-database.ts
LindvedKrvang: The instance variable `db` is optional, which is the variable parsed into this method.
The NonNullable allows us to write this:
```
public getCollection(collectionName: string): Collection {
  this.assertDatabaseConnection(this.db)
  return this.db.collection(collectionName)
}
```
Without the NonNullable we would also need a check to see if `this.db` is undefined or not.

src/model/interfaces/configured-input.ts
LindvedKrvang: I agree. The `Configured` is a weird name.
I have updated the names to follow your suggestion :) Hopefully I got all the places

src/model/interfaces/configured-panel-layout.ts
LindvedKrvang: Updated

src/data-access/interfaces/uuid-generator.ts
KvelaGorrrrnio: Nothing indicates that we are getting rid of dependency inversion – it is just food for thought :)

src/data-access/mongo/mongo-database.ts
KvelaGorrrrnio: By making the variable optional, it is assigned the type `Db | undefined`. The method return type states that if the method returns, then the argument given is of type `NonNullable<Db>` but since `Db` is not nullable, then `NonNullable<Db>` and `Db` are strictly equivalent. So the `NonNullable` wrapper is redundant.

I just tried removing NonNullable from the code and it works as intended.

Another suggestion: To make the use of `assertDatabaseConnection` a bit more straightforward, one could make the argument type `database: Db | undefined`. Having it as optional allows to call it like `this.assertDatabaseConnection()`, which will always fail. If the type is `database: Db | undefined`, then the user is forced to provide a parameter. This removes the confusion of whether the method works on internal data or not :)

src/model/interfaces/input-configuration.ts
```diff
@@ -0,0 +1,40 @@
...
+  onRelease?: boolean
+}
+
+export interface FaderConfiguration extends BaseInputConfiguration {
+  type: InputType.FADER
+}
+
+export type InputData = HttpEndpointInputData | ModifierInputData | TBarInputData
+
+export interface DefaultInputData {
```
KvelaGorrrrnio: It would be nice if this also used `Base` for consistency :)

KvelaGorrrrnio (approved):

src/data-access/mongo/mongo-database.ts
LindvedKrvang: I think I just understood what you were actually referring to about the `NonNullable` :P 
Then no, the `NonNullable` is not needed.

I have update it and also updated the parameter type :)

src/model/interfaces/input-configuration.ts
LindvedKrvang: So it becomes a `DataBase`? ;)

LindvedKrvang: Updated to `BaseInputData`

KvelaGorrrrnio: That would be nice and not very confusing.

MadsThy (approved):

KvelaGorrrrnio (approved):

PR #17, opened 2024-10-01
SOF-2229 Add basic endpoint validation

LindvedKrvang:
Update Update Update: Zod won. TypeBox implementation has been removed.
Now expose endpoint to create a PanelLayoutConfiguration. When it's create a PanelLayoutConfigurationCreated event is emitted.


src/presentation/controllers/base-controller.ts
```diff
@@ -22,16 +24,19 @@ interface Route {
   path: string
   method: Method
   action: Action
+  validationSchema: JSONSchema7 | undefined
```
MadsThy: Suggestion: I always default to using optionals in these cases 😄 What do you think?

```suggestion
  validationSchema?: JSONSchema7
```

src/presentation/controllers/panel-controller.ts
```diff
@@ -1,7 +1,19 @@
-import { BaseController, GetRequest, RestController } from './base-controller'
+import { BaseController, GetRequest, PostRequest, RestController } from './base-controller'
 import { FastifyReply, FastifyRequest } from 'fastify'
 import { PhysicalPanelLayoutRepository } from '../../data-access/physical-panel-layout-repository'
 import { PhysicalPanelLayout } from '../../model/interfaces/physical-panel-layout'
+import { JSONSchema7 } from 'json-schema'
+
+// TODO: This is for demonstration purposes. DELETE IT!
+const PANEL_LAYOUT_CONFIGURATION_SCHEMA: JSONSchema7 = {
```
MadsThy: From how i read this schema definition, i have to say I think I'm a bigger fan of Type-box. Mostly because I don't know if I see the need for having "required" properties :)

But other than that, I think I actually prefer JsonSchema over Zod. Haven't checked the performance numbers on the providers though... maybe we should look into that too :)

https://fastify.dev/docs/latest/Reference/Type-Providers/

MadsThy (dismissed):

src/presentation/controllers/base-controller.ts
LindvedKrvang: I'm also a bigger fan of the optional operator.
The reason I went for the `| undefined` in this case was because of this error:
image

If you have a nice suggestion for circumventing that, then I'm all for the optional operator :)

src/presentation/controllers/panel-controller.ts
LindvedKrvang: I will look into that :)

In this implementation, if we leave the `required` out, then we lose our validation. So for the current state, it is necessary.

LindvedKrvang: I have updated the PR to work with TypeBox.
I have kept the JSONSchema next to the TypeBox schema so people can see them side by side and evaluate themselves what they like.

Personally, I could be up for TypeBox :)

LindvedKrvang: In order to mark fields as required in TypeBox we would need to use the `Type.Required()`. 
So we won't be rid of the "required", but I think TypeBox does it more "elegant" than JSONSchema.

src/presentation/controllers/base-controller.ts
MadsThy: Oh right... Webstorm 😄 Haha... i still don't have a license for it.

Will check out what I can do to maybe get the same errors in VSCode, or just get a license 🥲

src/presentation/controllers/panel-controller.ts
```diff
@@ -1,7 +1,25 @@
...
+    hello: {
+      type: 'string',
+    },
+  },
+  required: ['hello'],
+}
+
+// TODO: This is for demonstration purposes. DELETE IT!
+const TYPE_BOX_PANEL_LAYOUT_CONFIGURATION_SCHEMA: TObject = Type.Object({
+  hello: Type.Required<TString>(Type.String()),
```
MadsThy: Holy fudge 😄 Yeah ok.. thats a gigantic difference in how to describe the schema... wow.. didn't realise it would look like this in the "real world".

I definitely would suggest going for TypeBox now :)

src/presentation/controllers/panel-controller.ts
LindvedKrvang: Nvm, `Type.Required()` is not required for the validation to work :)

MadsThy: Even better!!! :)

src/model/interfaces/input-configuration.ts
```diff
@@ -38,3 +40,95 @@ export interface ModifierInputData extends BaseInputData {
...
+  ZOD_MODIFIER_INPUT_DATA_SCHEMA,
+  ZOD_T_BAR_INPUT_DATA_SCHEMA,
+])
+
+const ZOD_BASE_INPUT_CONFIGURATION_SCHEMA: ZodType = z.object({
+  type: z.nativeEnum(InputType),
+  data: ZOD_INPUT_DATA_SCHEMA,
+})
+
+const ZOD_BUTTON_CONFIGURATION_SCHEMA: ZodType = z.intersection(ZOD_BASE_INPUT_CONFIGURATION_SCHEMA, z.object({
```
KvelaGorrrrnio: One thing I like about ZodType (or the alias ZodSchema), is that you can provide a type that the schema should uphold. So if the schema does not conform the the given type, then the type system will inform about it in the IDE or while transpiling.
E.g.:
image

KvelaGorrrrnio (commented):
I think the choice comes down to 3 things:
- Performance.
- Compliance (e.g. with OpenAPI and JSON Schema).
- Data modelling capabilities.

In regards to performance, then TypeBox is faster, since it only validates the data. I don't think that the performance difference between TypeBox and Zod is something that will affect us though.

Compliance with OpenAPI and JSON schema is something we should have a chat about. If it something we favor, then TypeBox is a good choice, though there also exists Zod-to-JSON-schema libraries.

I am not a big fan of the use of `maxProperties` to distinguish between different schemas to model XOR. If it is not possible to model a type union of complex data correctly in TypeBox, then that is a big turn-off for me.
So if we don't care about compliance, then I am in favor of Zod :)

src/model/interfaces/input-configuration.ts
LindvedKrvang: That was the type I couldn't find. I have updated to use `ZodSchema` with the proper interfaces.

MadsThy (dismissed):

RasmusAlbrektsen (dismissed):

PR #23, opened 2024-10-07
SOF-2230 Set up initial EventEmitter system

LindvedKrvang:
Initial EventEmitter system has been set up. It follows the EventEmitter system used in SofieServer.
Currently it only supports emitting a `PanelLayoutConfigurationCreatedEvent`. Other events should just be added as needed.

No longer use the `wsUpstream` provided by Fasitfy. This is so we can emit the events from SofieTv2Server on the same port as the events being parsed from SofieServer.
Instead SofieTv2Server exposes a WebSocketServer through Fastify that clients connects to. SofieTv2Server now also creates a WebSocket Client that connects to SofieServer and all Events received on this client, is being emitted to all clients listening to Fastifys WebSocketServer. (This should also make it possible for us to intercept the events from SofieServer and manipulate them if we want to.)

The WebSocket Client only has a very simple reconnect strategy, which is to simply to connect again after five seconds. We would need to implement a more robust WebSocket.

Overall, we need to do something about the `FastifyServer`. It's starting to accumulate responsibilities.


package.json
```diff
@@ -26,6 +26,7 @@
     "@stylistic/eslint-plugin": "^1.8.0",
     "@types/license-checker": "^25",
     "@types/node": "^20.12.8",
+    "@types/ws": "^8.5.12",
```
KvelaGorrrrnio, line 29: If we depend on the package 'ws', then we should put in the dependencies section as well :)
Note: There is also a WebSocket client implementation in Node. In Node version 20 it can be enabled with the flag `--experimental-websocket`. From Node version 22, it is available by default.

src/presentation/fastify-server.ts
```diff
@@ -5,17 +5,24 @@ import { ProxyConfiguration } from './value-objects/proxy-configuration'
 import { BaseController } from './controllers/base-controller'
 import { RouteOptions } from 'fastify/types/route'
 import { Logger } from '../logger/logger'
+import { fastifyWebsocket } from '@fastify/websocket'
+import WebSocket, { MessageEvent } from 'ws'
+import { PanelEventObserver } from './interfaces/panel-event-observer'
+import { PanelEvent } from './value-objects/panel-event'
+
+const RECONNECT_TO_ALBA_SERVER_IN_MS: number = 5_000
```
KvelaGorrrrnio: Just a note: This describes what server we are connecting to/where it used :) Nothing in this file indicates that it is an Alba backend we connect to.
You can consider renaming it something like `RECONNECT_DELAY_IN_MS`.

src/presentation/fastify-server.ts
```diff
@@ -25,8 +32,29 @@ export class FastifyServer implements ProxyServer {
...
+  private async setupWebSocketServer(proxyConfiguration: ProxyConfiguration): Promise<void> {
+    await this.fastifyServer.register(fastifyWebsocket)
+    this.fastifyServer.get('/ws', { websocket: true }, (socket: WebSocket) => {
+      this.panelEventObserver.subscribeToPanelEvents((panelEvent: PanelEvent) => socket.send(JSON.stringify(panelEvent)))
+    })
+
+    this.connectToAlbServer(proxyConfiguration)
+  }
+
+  private connectToAlbServer(proxyConfiguration: ProxyConfiguration): void {
```
KvelaGorrrrnio: `Alb` is missing an `a`, but is `Alba` needed here?

src/presentation/fastify-server.ts
```diff
@@ -25,8 +32,29 @@ export class FastifyServer implements ProxyServer {
...
+    const albaWebSocket: WebSocket = new WebSocket(proxyConfiguration.websocketUrl)
+    albaWebSocket.addEventListener('open', () => this.logger.debug('Connected to AlbaServer'))
+
+    albaWebSocket.addEventListener('close', () => {
+      this.logger.debug('Disconnected from Alba Server')
+      setTimeout(() => this.connectToAlbServer(proxyConfiguration), RECONNECT_TO_ALBA_SERVER_IN_MS)
```
KvelaGorrrrnio: If/when the connection is closed, we should think about how we propagate that information to the frontend. Either we drop all connections to the websocket server, or emit some event to the frontend, or silently hope that the clients don't realize that we have down-time until we have re-established a connection.

src/presentation/fastify-server.ts
```diff
@@ -25,8 +32,29 @@ export class FastifyServer implements ProxyServer {
...
+    })
+  }
+
+  private async setupWebSocketServer(proxyConfiguration: ProxyConfiguration): Promise<void> {
+    await this.fastifyServer.register(fastifyWebsocket)
+    this.fastifyServer.get('/ws', { websocket: true }, (socket: WebSocket) => {
+      this.panelEventObserver.subscribeToPanelEvents((panelEvent: PanelEvent) => socket.send(JSON.stringify(panelEvent)))
+    })
+
+    this.connectToAlbServer(proxyConfiguration)
```
KvelaGorrrrnio: I have no problem with this, but has it been given thought to whether this method should await the connection to the server to be established?

KvelaGorrrrnio (approved):
Looks good :rocket:

src/presentation/fastify-server.ts
LindvedKrvang: Not much though has been given. This is mainly just a simply, not robust, implementation. The idea is to have a seperate task to make it proper.

src/presentation/fastify-server.ts
LindvedKrvang: Agreed, the though was to do that in a seperate task

MadsThy (approved):

src/presentation/fastify-server.ts
LindvedKrvang: Maybe it is not. I would like however, to have some way to differentiate that this method is not meant to connect to the FastifyServer (which is the class we are in).

I have added the `a` for now.

src/presentation/fastify-server.ts
LindvedKrvang: Updated

package.json
LindvedKrvang: Dependency added

RasmusAlbrektsen (approved):

PR #30, opened 2024-10-14
SOF-2233 Expose endpoint to Update PanelLayoutConfiguration

LindvedKrvang:
Note: Dependent on https://github.com/tv2/sofie-tv2-server/pull/29

Expose endpoint to Update a PanelLayoutConfiguration.
When a PanelLayoutConfiguration is updated, an update event is also emitted.

Also in this PR: Update the `PutRequest` annotation to also take a Schema for validation of the body of the Request.


src/data-access/mongo/mongo-panel-layout-configuration-repository.ts
```diff
@@ -42,6 +42,11 @@ export class MongoPanelLayoutConfigurationRepository extends BaseMongoRepository
     return panelLayoutConfigurationToBeInserted
   }
 
+  public async updatePanelLayoutConfiguration(panelLayoutConfiguration: PanelLayoutConfiguration): Promise<void> {
+    this.assertDatabaseConnection(this.updatePanelLayoutConfiguration.name)
+    await this.getCollection().updateOne({ id: panelLayoutConfiguration.id }, { $set: panelLayoutConfiguration })
```
MadsThy, line 47: What is the $set for? :O Is that some kind of mongo thing I'm just not aware of? :) Never seen anyone name things with $ in front

MadsThy (dismissed):

src/presentation/controllers/panel-controller.ts
```diff
@@ -63,6 +64,17 @@ export class PanelController extends BaseController {
...
   }
 
+  @PutRequest('/panelLayoutConfigurations', ZOD_PANEL_LAYOUT_CONFIGURATION_SCHEMA)
+  public async updatePanelLayoutConfiguration(request: FastifyRequest<{ Body: PanelLayoutConfiguration }>, reply: FastifyReply): Promise<void> {
+    try {
+      const panelLayoutConfiguration: PanelLayoutConfiguration = request.body
+      await this.panelLayoutConfigurationService.updatePanelLayoutConfiguration(panelLayoutConfiguration)
+      await reply.code(HttpStatusCode.OK).send(`Successfully updated PanelLayoutConfiguration for ${panelLayoutConfiguration.id}`)
+    } catch (error) {
+      await this.httpErrorHandler.handleError(reply, error as Exception)
```
KvelaGorrrrnio: Unrelated thought. It would be nice if the `HttpErrorHandler.handleError` took an unknown instead and the it could check if it is an Exception or Error. In the case of an unknown type, then 500 could be used as fallback.

KvelaGorrrrnio (dismissed):

src/data-access/mongo/mongo-panel-layout-configuration-repository.ts
RasmusAlbrektsen: It's a Mongo query operator: https://www.mongodb.com/docs/manual/reference/operator/update/set/ :-)

RasmusAlbrektsen (dismissed):

src/data-access/mongo/mongo-panel-layout-configuration-repository.ts
LindvedKrvang: As Rasmus has already stated, it's a MongoDb operator. If you don't use it you would get an error complaining about the transaction not being an Atomic operation.

RasmusAlbrektsen (approved):

src/presentation/controllers/panel-controller.ts
LindvedKrvang: Added

LindvedKrvang: Also added a `HttpResponseFormatter` so that we can format after Jsend

KvelaGorrrrnio (approved):

PR #31, opened 2024-10-14
Sof 2236/expose create panel configuration endpoint

LindvedKrvang:
Note: Dependent on PR: https://github.com/tv2/sofie-tv2-server/pull/30

Expose endpoint to create a PanelConfiguration.
When a PanelConfiguration is created a PanelConfigurationCreated event is emitted.

It is not possible to create a PanelConfiguration that either don't have a `panelLayoutConfigurationId` for a valid `panelLayoutConfiguration` or if the `model` and `type` of the PanelConfiguration does not match the `model` and `type` of the `PanelLayoutConfiguration`.

Also in this PR: Added missing tests for testing whether events are emitted for CRUD operations of PanelLayoutConfiguration.


src/business-logic/services/panel-service-implementation.ts
```diff
@@ -29,4 +37,27 @@ export class PanelServiceImplementation implements PanelService {
...
+      }
+      throw new UnsupportedOperationException(`Can't create PanelConfiguration. No PanelLayoutConfiguration exist for ${panelConfigurationWithoutId.panelLayoutConfigurationId}`)
+    }
+
+    if (panelLayoutConfiguration.model !== panelConfigurationWithoutId.model || panelLayoutConfiguration.type !== panelConfigurationWithoutId.type) {
+      throw new UnsupportedOperationException(
+        'Can\'t create PanelConfiguration. Both PanelModel and PanelType needs to match on PanelConfiguration and PanelLayoutConfiguration.'
+        + `PanelConfiguration: Model ${panelConfigurationWithoutId.model}, Type: ${panelConfigurationWithoutId.type} - `
+        + `PanelLayoutConfiguration: Model: ${panelLayoutConfiguration.model}, Type: ${panelLayoutConfiguration.type}`
+      )
```
MadsThy, line 57:
```suggestion
      throw new UnsupportedOperationException(
  `Can't create PanelConfiguration. Both PanelModel and PanelType need to match on PanelConfiguration and PanelLayoutConfiguration.
  PanelConfiguration: Model ${panelConfigurationWithoutId.model}, Type: ${panelConfigurationWithoutId.type} - 
  PanelLayoutConfiguration: Model: ${panelLayoutConfiguration.model}, Type: ${panelLayoutConfiguration.type}`
)
```

src/data-access/mongo/mongo-panel-configuration-repository.ts
```diff
@@ -0,0 +1,29 @@
+import { PanelConfigurationRepository } from '../interfaces/panel-configuration-repository'
+import { BaseMongoRepository } from './base-mongo-repository'
+import { MongoDatabase } from './mongo-database'
+import { UuidGenerator } from '../interfaces/uuid-generator'
+import { util } from 'zod'
+import Omit = util.Omit
```
MadsThy:
```suggestion
import { util } from 'zod'
import Omit = util.Omit
```

Do you really need these? :) Do you need to get the Omit type from zod? Cause when I delete those two imports, the file still checks out fine.

src/presentation/controllers/panel-controller.ts
```diff
@@ -57,8 +61,8 @@ export class PanelController extends BaseController {
   public async createPanelLayoutConfiguration(request: FastifyRequest<{ Body: PanelLayoutConfiguration }>, reply: FastifyReply): Promise<void> {
     try {
       const panelLayoutConfiguration: PanelLayoutConfiguration = request.body
-      await this.panelLayoutConfigurationService.createPanelLayoutConfiguration(panelLayoutConfiguration)
-      await reply.code(HttpStatusCode.OK).send(panelLayoutConfiguration)
+      await this.panelService.createPanelLayoutConfiguration(panelLayoutConfiguration)
+      await reply.code(HttpStatusCode.OK).send('Successfully create PanelLayoutConfiguration')
```
MadsThy:
```suggestion
      await reply.code(HttpStatusCode.OK).send('Successfully created PanelLayoutConfiguration')
```

MadsThy (dismissed):

src/presentation/controllers/panel-controller.ts
KvelaGorrrrnio: Just a note: Alba server uses Jsend for formatting the response. It would make sense to have the same format here.

src/data-access/mongo/mongo-panel-configuration-repository.ts
```diff
@@ -0,0 +1,29 @@
...
+    return PANEL_CONFIGURATION_COLLECTION_NAME
+  }
+
+  public async createPanelConfiguration(panelConfigurationWithoutId: Omit<PanelConfiguration, 'id'>): Promise<PanelConfiguration> {
+    this.assertDatabaseConnection(this.createPanelConfiguration.name)
+    const panelConfigurationToBeInserted: PanelConfiguration = {
+      ...panelConfigurationWithoutId,
+      id: this.uuidGenerator.generateUuid()
+    }
+    await this.getCollection().updateOne({ id: panelConfigurationToBeInserted.id }, { $set: panelConfigurationToBeInserted }, { upsert: true })
```
KvelaGorrrrnio, line 24: Is there a reason that `updateOne` is used over `insertOne` here?

package.json
```diff
@@ -24,17 +24,20 @@
   },
   "devDependencies": {
     "@stylistic/eslint-plugin": "^1.8.0",
+    "@types/jest": "^29.5.13",
```
KvelaGorrrrnio: Why is `@types/jest` and `ts-jest` added?

src/business-logic/services/panel-service-implementation.ts
```diff
@@ -29,4 +37,27 @@ export class PanelServiceImplementation implements PanelService {
...
+  public async createPanelConfiguration(panelConfigurationWithoutId: PanelConfiguration): Promise<void> {
+    let panelLayoutConfiguration: PanelLayoutConfiguration
+    try {
+      panelLayoutConfiguration = await this.panelLayoutConfigurationRepository.getPanelLayoutConfiguration(panelConfigurationWithoutId.panelLayoutConfigurationId)
+    } catch (error) {
+      if (!(error instanceof NotFoundException)) {
+        throw error
+      }
+      throw new UnsupportedOperationException(`Can't create PanelConfiguration. No PanelLayoutConfiguration exist for ${panelConfigurationWithoutId.panelLayoutConfigurationId}`)
+    }
```
KvelaGorrrrnio, line 50: Moving this to a separate method will get rid og the `let` binding and will make this method more manageable.

RasmusAlbrektsen (dismissed):

src/business-logic/services/panel-service-implementation.ts
LindvedKrvang: Wouldn't this give you a lot of `\n` in your string? 
That's atleast what I got before I seperated them into sperate strings

src/business-logic/services/panel-service-implementation.ts
LindvedKrvang: Which part of it are you referring to?

src/presentation/controllers/panel-controller.ts
LindvedKrvang: It has been added in https://github.com/tv2/sofie-tv2-server/pull/30

src/data-access/mongo/mongo-panel-configuration-repository.ts
LindvedKrvang: Yes, `insertOne` requires that the entity we are inserting to have `_id` which we don't want

package.json
LindvedKrvang: To use for test

src/data-access/mongo/mongo-panel-configuration-repository.ts
LindvedKrvang: No, I think that was auto-import gone wrong. Very nice catch :D

src/presentation/controllers/panel-controller.ts
LindvedKrvang: Updated

package.json
KvelaGorrrrnio: But the repo is not set up to use jest - it uses vitest. I tried removing these dependencies and the tests still run :)

src/business-logic/services/panel-service-implementation.ts
KvelaGorrrrnio: Never mind - I missed what the error message says :)

KvelaGorrrrnio (approved):
Under the assumption that the jest dependencies are removed, there is a green light from me 🟢

KvelaGorrrrnio (dismissed):

PR #35, opened 2024-10-18
Sof 2241/expose endpoint to update panel configuration

LindvedKrvang:
Note: Dependent on https://github.com/tv2/sofie-tv2-server/pull/34
Update: All dependent PRs have been merged

Expose endpoint to update a `PanelConfiguration`.
If the PanelConfiguration to be updated has a different `model` or `type` than the associated `PanelLayoutConfiguration` then the update fails. The same happens if the `PanelLayoutConfiguration` that has different `model` or `type`.

Also in this PR:
### A new approach to write test-setup
The goal with this setup is to minimize the amount of boilerplate code needed for each test, while ensuring that test-setup doesn't "bleed" into other tests.

The basic idea is that since each _describe_ tells a scenario, that scenario would then be setup in all tests under said _describe_.
The basic concept would look like this:
```
describe(CarPainter.name, () => {
  let car: Car

  describe('the Car is red', () => {
    beforeEach(() => {
      car = new Car(Color.RED) // Creates a Car that always has the color Red.
    })

    it('paints the Car blue', () => {
      const testee: CarPainter = new CarPainter()
      testeee.paintCar(car, Color.BLUE)
      expect(car.color).toBe(Color.BLUE)
    }
 
    ...
  }

  describe('the Car is green', () => {
    beforeEach(() => {
      car = new Car(Color.GREEN) // Creates a Car that always has the color Green.
    })

    ...
  }
}
```
**Pros:**
- Tests that have the same test-setup doesn't need to repeat the same test-setup over and over again which in turn should reduce the amount of boilerplate code needed.

**Cons:**
- It can potentially be difficult to get an overview of the actual test-setup since it can be spread of over multiple _describe_
- - The theory is: If done right, one can simply look at the describes and know what the test-setup is.

Disclaimer: I don't have any practical experience with this approach, so it's currently only on theory basis. So the idea was to try it out and see if it is something we want. I would very much like your opinions on this approach and whether you think it is something we might adopt.

Update:
I have tried applying the new test-setup structure to its fullest in this PR: https://github.com/tv2/sofie-tv2-server/pull/37
A more clean view of the entire file can be found here: https://github.com/tv2/sofie-tv2-server/blob/SOF-2305/guardUpdateOfPanelLayoutConfiguration/src/business-logic/services/test/panel-service-implementation.spec.ts


src/business-logic/services/panel-service-implementation.ts
```diff
@@ -56,25 +56,43 @@ export class PanelServiceImplementation implements PanelService {
...
+
+    if (panelLayoutConfiguration.model !== panelConfigurationWithoutId.model || panelLayoutConfiguration.type !== panelConfigurationWithoutId.type) {
+      this.throwUnsupportedOperationWithIncompatibleMessage(panelConfigurationWithoutId, panelLayoutConfiguration, 'create')
+    }
+
+    const panelConfiguration: PanelConfiguration = await this.panelConfigurationRepository.createPanelConfiguration(panelConfigurationWithoutId)
+    this.panelEventEmitter.emitPanelConfigurationCreated(panelConfiguration)
+  }
+
+  private async fetchPanelLayoutConfiguration(panelLayoutConfigurationId: string, verb: string): Promise<PanelLayoutConfiguration> {
```
RasmusAlbrektsen: Genuinely curious, Is `verb` a normal approach to handling this? 
Other than something like, perhaps, `action` or `intention` (just throwing examples off the top of my head) :-)

RasmusAlbrektsen (approved):
I like the test-approach, especially how we `describe` a scenario, and subsequently set it up right after in the `beforeEach`. It also makes a failing `it` easier to understand, since the general scenario setup should be something we could just read and assume is correct before we make it to this point.

src/business-logic/services/panel-service-implementation.ts
LindvedKrvang: Don't think it is. `verb` is something I picked at "semi" random :p 
Suggestions are much appreciated. I'm not a fan of currently writing the magic string whenever we use the method.

RasmusAlbrektsen: I guess we could have something like an enum that correlates to the type of request as reference? That's what i got for now at least :-D

KvelaGorrrrnio: `operation` could also be an alternative :)

src/business-logic/services/panel-service-implementation.ts
```diff
@@ -56,25 +56,43 @@ export class PanelServiceImplementation implements PanelService {
...
-        'Can\'t create PanelConfiguration. Both PanelModel and PanelType needs to match on PanelConfiguration and PanelLayoutConfiguration.'
-        + `PanelConfiguration: Model ${panelConfigurationWithoutId.model}, Type: ${panelConfigurationWithoutId.type} - `
-        + `PanelLayoutConfiguration: Model: ${panelLayoutConfiguration.model}, Type: ${panelLayoutConfiguration.type}`
-      )
+  private throwUnsupportedOperationWithIncompatibleMessage(panelConfiguration: PanelConfiguration, panelLayoutConfiguration: PanelLayoutConfiguration, verb: string): void {
+    throw new UnsupportedOperationException(
+      `Can't ${verb} PanelConfiguration. Both PanelModel and PanelType needs to match on PanelConfiguration and PanelLayoutConfiguration.`
+      + `PanelConfiguration: Model ${panelConfiguration.model}, Type: ${panelConfiguration.type} - `
+      + `PanelLayoutConfiguration: Model: ${panelLayoutConfiguration.model}, Type: ${panelLayoutConfiguration.type}`
+    )
```
KvelaGorrrrnio: Who is this message intended for? If the intention is that this is exposed to users/clients, then I would argue that it might have to much information and is formatted in a too technical way.
If it is intended for logging purposes, then a more specialized exception could be used, that takes the panel layout configuration and panel configuration as arguments. That way, the models can be retrieved for better log structuring.
E.g.
```typescript
  throw new IncompatiblePanelAndPanelLayoutException(`Unable to ${operation} panel, since panel model and/or panel type are incompatible with the chosen panel layout.`, panelConfiguration, panelLayoutConfiguration)
```

src/business-logic/services/panel-service-implementation.ts
```diff
@@ -56,25 +56,43 @@ export class PanelServiceImplementation implements PanelService {
...
+  }
+
+  public async updatePanelConfiguration(panelConfiguration: PanelConfiguration): Promise<void> {
+    const panelLayoutConfiguration: PanelLayoutConfiguration = await this.fetchPanelLayoutConfiguration(panelConfiguration.panelLayoutConfigurationId, 'update')
+
+    if (panelConfiguration.model !== panelLayoutConfiguration.model || panelConfiguration.type !== panelLayoutConfiguration.type) {
+      this.throwUnsupportedOperationWithIncompatibleMessage(panelConfiguration, panelLayoutConfiguration, 'update')
     }
```
KvelaGorrrrnio: Suggestion: Instead of having the check wherever we want to throw the exception, the condition could be moved to a method (e.g. merged into `throwUnsupportedOperationWithIncompatibleMessage`), so this method could simply call something like 
```typescript
this.assertPanelAndPanelLayoutCompatibility(panelConfiguration, panelLayoutConfiguration, 'update').
```

src/business-logic/services/panel-service-implementation.ts
```diff
@@ -56,25 +56,43 @@ export class PanelServiceImplementation implements PanelService {
   }
 
   public async createPanelConfiguration(panelConfigurationWithoutId: PanelConfiguration): Promise<void> {
-    let panelLayoutConfiguration: PanelLayoutConfiguration
+    const panelLayoutConfiguration: PanelLayoutConfiguration = await this.fetchPanelLayoutConfiguration(panelConfigurationWithoutId.panelLayoutConfigurationId, 'create')
+
+    if (panelLayoutConfiguration.model !== panelConfigurationWithoutId.model || panelLayoutConfiguration.type !== panelConfigurationWithoutId.type) {
+      this.throwUnsupportedOperationWithIncompatibleMessage(panelConfigurationWithoutId, panelLayoutConfiguration, 'create')
+    }
+
```
KvelaGorrrrnio: Should it be mandatory to assign a layout to a panel?
From a user perspective it gives more flexibility to be able to create all panels that should be connected before creating the layouts or vice versa. It will also make it easier to e.g. upload a set of panel configurations from e.g. a JSON file.

KvelaGorrrrnio (approved):

src/business-logic/services/panel-service-implementation.ts
LindvedKrvang: It is intended for the user, since this will be thrown after a user action.
What format/message do you suggest instead?

src/business-logic/services/panel-service-implementation.ts
LindvedKrvang: That is a PO question.
My only opinion on the matter right now is that it would be nice with one less optional attribute.

src/business-logic/services/panel-service-implementation.ts
LindvedKrvang: I have introduced `assertPanelLayoutConfigurationAndPanelConfigurationCompatibility(...)`

src/business-logic/services/panel-service-implementation.ts
LindvedKrvang: Update to `operation`.

src/business-logic/services/panel-service-implementation.ts
KvelaGorrrrnio: The team is the PO (to some degree) ;)
But we can look into it at another point in time.

src/business-logic/services/panel-service-implementation.ts
KvelaGorrrrnio: Then I would suggest something like what I wrote in the snippet:
`Unable to ${operation} panel, since panel model and/or panel type are incompatible with the chosen panel layout.`

KvelaGorrrrnio (approved):

src/business-logic/services/panel-service-implementation.ts
LindvedKrvang: Updated the message to your suggestion.
Also removed the method and moved the Exception up into `assertPanelLayoutConfigurationAndPanelConfigurationCompatibility` since that's now the only one calling it.

src/business-logic/services/panel-service-implementation.ts
```diff
@@ -65,7 +65,7 @@ export class PanelServiceImplementation implements PanelService {
 
   private assertPanelLayoutConfigurationAndPanelConfigurationCompatibility(panelLayoutConfiguration: PanelLayoutConfiguration, panelConfiguration: PanelConfiguration, operation: string): void {
     if (panelLayoutConfiguration.model !== panelConfiguration.model || panelLayoutConfiguration.type !== panelConfiguration.type) {
-      this.throwUnsupportedOperationWithIncompatibleMessage(panelConfiguration, panelLayoutConfiguration, operation)
+      throw new UnsupportedOperationException(`Unable to ${operation} panel, since panel model and/or panel type are incompatible with the chosen panel layout`)
```
KvelaGorrrrnio, line 68: It would be nice with a `.` at the end of sentences.

KvelaGorrrrnio (approved):

PR #36, opened 2024-10-18
SOF-2242 Expose delete endpoint for PanelConfiguration

LindvedKrvang:
Note: Dependent on https://github.com/tv2/sofie-tv2-server/pull/35

Expose an endpoint to delete a PanelConfiguration.


RasmusAlbrektsen (approved):

KvelaGorrrrnio (approved):
Looks good! :) I am still unsure about the new approach for structuring test setups, but we can try it out in this repo and see if it is something we should continue doing.

MadsThy (approved):

PR #37, opened 2024-10-18
Sof 2305/guard update of panel layout configuration

LindvedKrvang:
Note: Dependent on https://github.com/tv2/sofie-tv2-server/pull/36

No longer possible to update a `PanelLayoutConfiguration` if a `PanelConfiguration` is using it.

Also in this PR:
Tried taking `panel-service-implementation.spec.ts` to the "extreme" of the new test-setup approach


KvelaGorrrrnio (commented):
Why is it that we don't want to update an assigned layout?
I might be wrong, but isn't it only the `model` and `type` of the layout that should be immutable?

MadsThy (approved):

LindvedKrvang:
> Why is it that we don't want to update an assigned layout? I might be wrong, but isn't it only the `model` and `type` of the layout that should be immutable?

If I am honest, I think I was too lazy and this was easy :p 
If consensus is to only let `model` and `type` be immutable, then I will update it :)

src/model/interfaces/panel-layout-configuration.ts
```diff
@@ -17,6 +17,8 @@ export const ZOD_PANEL_LAYOUT_CONFIGURATION_SCHEMA_WITHOUT_ID: ZodSchema<Omit<Pa
...
 })
 
-export const ZOD_PANEL_LAYOUT_CONFIGURATION_SCHEMA: ZodSchema<PanelLayoutConfiguration> = ZOD_PANEL_LAYOUT_CONFIGURATION_SCHEMA_WITHOUT_ID.and(z.object({
-  id: z.string()
-}))
+export const ZOD_UPDATE_PANEL_LAYOUT_CONFIGURATION_SCHEMA: ZodSchema<Omit<PanelLayoutConfiguration, 'model' | 'type'>> = z.object({
+  id: z.string(),
+  name: z.string(),
+  inputConfigurations: z.record(z.string(), ZOD_INPUT_CONFIGURATION_SCHEMA),
+})
```
KvelaGorrrrnio, line 24: Should we use strict matching, so an error is thrown if `model` and/or `type` is provided. It will make it obvious if we accidentally tries to modify the attributes.

src/business-logic/services/panel-service-implementation.ts
```diff
@@ -34,16 +34,16 @@ export class PanelServiceImplementation implements PanelService {
...
     this.panelEventEmitter.emitPanelLayoutConfigurationDeleted(panelLayoutConfigurationId)
   }
 
-  private async assertNoPanelConfigurationIsUsingThePanelLayoutConfiguration(panelLayoutConfigurationId: string): Promise<void> {
+  private async assertNoPanelConfigurationIsUsingThePanelLayoutConfiguration(panelLayoutConfigurationId: string, operation: string): Promise<void> {
     const panelConfigurationsForPanelLayoutConfiguration: PanelConfiguration[] = await this.panelConfigurationRepository.getPanelConfigurationsForPanelLayoutConfiguration(panelLayoutConfigurationId)
     if (panelConfigurationsForPanelLayoutConfiguration.length > 0) {
       const panelConfigurationIds: string[] = panelConfigurationsForPanelLayoutConfiguration.map(panelConfiguration => panelConfiguration.id)
-      throw new UnsupportedOperationException(`Can't delete PanelLayoutConfiguration for ${panelLayoutConfigurationId}. It's in use by the ConfigurationPanels [${panelConfigurationIds}]`)
+      throw new UnsupportedOperationException(`Can't ${operation} PanelLayoutConfiguration for ${panelLayoutConfigurationId}. It's in use by the ConfigurationPanels [${panelConfigurationIds}]`)
```
KvelaGorrrrnio: I assume that this message is user facing, so I suggest that hostnames are used instead of ids and that we don't use class/interface names for entities:
```typescript
throw new UnsupportedOperationException(`Unable to ${operation} panel layout '${panelLayout.name}' with id '${panelLayout.id}', since it is used by the panel(s): ${panelConfigurationsForPanelLayoutConfiguration.map(panelConfiguration => panelConfiguration.hostname).join(', ')}.`)
```

KvelaGorrrrnio (approved):

src/model/interfaces/panel-layout-configuration.ts
LindvedKrvang: I don't have a big preference. How do you suggest to to do that?

src/business-logic/services/panel-service-implementation.ts
LindvedKrvang: Updated to show the names of the PanelConfigurations instead of the id. For the `PanelLayoutConfiguration`, all we have is the id, so it's difficult to show any but the id.

PR #38, opened 2024-10-18
Sof 2247/connect to panels configured on startup

LindvedKrvang:
Note: Dependent on https://github.com/tv2/sofie-tv2-server/pull/37

On startup, we now load all PanelConfigurations and connect to their Panel (as long as it is a valid Skaarhoj Panel that has been configured).
If we connect, we emit a StatusMessage event saying we connected to the device. If we fail to connect, we emit an event saying that we failed.

Also in this PR:
- Events for StatusMessages
- "Saving" StatusMessages in-memory. The idea is that they are derived from the current state, so no need to persist them.
- "Hook" into the `getStatusMessages` endpoint so that clients gets StatusMessages from both AlbaServer and AlbaTv2Server with a single call.
- Able to make HttpRequest from the logic layer to other services (currently used to call AlbaServer).
- Moved some misplaced files around


KvelaGorrrrnio:
One of the reasons that the placement of the control panel integration was chosen to be in this component, was that status management could be handled locally (so it wouldn't need to be send it to Alba server).

LindvedKrvang:
> One of the reasons that the placement of the control panel integration was chosen to be in this component, was that status management could be handled locally (so it wouldn't need to be send it to Alba server).

The main reason the integration was chosen to be handled in this repo is to keep Alba servers responsibilities to a minimum and mainly focus on calculating Timelines etc

KvelaGorrrrnio:
> The main reason the integration was chosen to be handled in this repo is to keep Alba servers responsibilities to a minimum and mainly focus on calculating Timelines etc

The choice was more between having it as a separate application/repo (like sofie-skaarhoj-poc) vs having it in this repo. So neither of these would impose extra responsibilities on alba-server.

src/business-logic/services/interfaces/panel.ts
```diff
@@ -0,0 +1,3 @@
+export interface Panel {
```
MadsThy, line 1: When i see this interface, i feel like I want us to add this ESLint rule 😄 

`"@typescript-eslint/no-empty-object-type": "error"`

src/data-access/in-memory/in-memory-status-message-repository.ts
```diff
@@ -0,0 +1,30 @@
...
+  public static getInstance(): StatusMessageRepository {
+    if (!this.instance) {
+      this.instance = new InMemoryStatusMessageRepository()
+    }
+    return this.instance
+  }
+
+  private readonly statusMessages: Map<string, StatusMessage> = new Map()
+
+  private constructor() { }
```
MadsThy, line 17: This one makes me want to introduce these rules 😄 

`
"no-useless-constructor": "off",
"@typescript-eslint/no-useless-constructor": "error"
`

src/model/entities/status-message.ts
```diff
@@ -0,0 +1,9 @@
+import { StatusCode } from '../enums/status-code'
+
+export interface StatusMessage {
+  id: string
+  statusCode: StatusCode
+  title: string
+  message: string
+  lastUpdatedTimestamp?: number
```
MadsThy: Just cause OCD i guess 😆 

```suggestion
  id: string
  title: string
  message: string
  statusCode: StatusCode
  lastUpdatedTimestamp?: number
```

src/business-logic/panel-integrations/skaarhoj-panel.ts
```diff
@@ -0,0 +1,75 @@
...
+export class SkaarhojPanel implements Panel {
+  private readonly logger: Logger
+  private socket: Socket = new Socket()
+
+  public constructor(
+    private readonly panelConfiguration: PanelConfiguration,
+    private readonly statusMessageService: StatusMessageService,
+    logger: Logger
+  ) {
+    this.logger = logger.tag(`${SkaarhojPanel.name}:${panelConfiguration.hostname}`)
```
KvelaGorrrrnio, line 22: Note that if the hostname panelConfiguration changes, then this will be outdated.

src/business-logic/panel-integrations/skaarhoj-panel.ts
```diff
@@ -0,0 +1,75 @@
...
+    this.socket.on('data', (data) => {
+      this.logger.data(data).info(`Received input from ${SkaarhojPanel.name}:${this.panelConfiguration.hostname}`)
+    })
+  }
+
+  private createConnectionSuccessStatusMessage(): StatusMessage {
+    return {
+      id: this.getStatusMessageId(),
+      title: 'Connection complete',
+      message: `Successfully connected to Skaarhoj Panel on: ${this.panelConfiguration.hostname}`,
```
KvelaGorrrrnio: To make this a bit more user friendly, I would remove the `:` and add `.` at the end of the sentence. The user doesn't need to know of the code construct `Panel`, so it can be written with a lowercase p. And I suggest that `on` is replaced with `at`. Is there a case where we unsuccessfully connects to a panel or can `Successfully`.
Combining it, I will suggest to change message to the `Connected to the Skaarhoj panel at ${this.panelConfiguration.hostname}.`

src/business-logic/panel-integrations/skaarhoj-panel.ts
```diff
@@ -0,0 +1,75 @@
...
+
+  private getStatusMessageId(): string {
+    return `skaarhoj_${this.panelConfiguration.hostname}`
+  }
+
+  private createUnableToConnectStatusMessage(): StatusMessage {
+    return {
+      id: this.getStatusMessageId(),
+      title: 'Unable to connect',
+      message: `Unable to connect to the Skaarhoj Panel on: ${this.panelConfiguration.hostname}`,
```
KvelaGorrrrnio: The title should at a glance give away the context. `Unable to connect` could mean a lot of things in the system. I suggest that it is changed to `Unreachable Skaarhoj panel` or `Skaarhoj panel is unreachable`.

And since the message is a description for the user I will make the same suggestion as for comment under `createConnectionSuccessStatusMessage`, so my suggestion would be that the message is something like `Unable to connect to the Skaarhoj panel at ${this.panelConfiguration.hostname}.`

We can consider adding a user friendly name to the panel configurations and use that instead of the hostname/IP.

src/business-logic/panel-integrations/skaarhoj-panel.ts
```diff
@@ -0,0 +1,75 @@
...
+    this.socket.setEncoding('utf8')
+    this.socket.on('data', (data) => {
+      this.logger.data(data).info(`Received input from ${SkaarhojPanel.name}:${this.panelConfiguration.hostname}`)
+    })
+  }
+
+  private createConnectionSuccessStatusMessage(): StatusMessage {
+    return {
+      id: this.getStatusMessageId(),
+      title: 'Connection complete',
```
KvelaGorrrrnio: `Connection complete` can mean a lot of things in the system. To give the user a bit more information at a glance, I suggest that the title is something like `Connected to Skaarhoj Panel`.

src/business-logic/panel-integrations/test/skaarhoj-panel.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+    })
+
+    describe('it receives a PanelConfiguration with Skaarhoj type and model', () => {
+      it('does not throw any exception', () => {
+        panelConfiguration = EntityTestFactory.createPanelConfiguration({ type: PanelType.SKAARHOJ, model: SkaarhojModel.MK48 })
+        try {
+          new SkaarhojPanel(panelConfiguration, instance(mock(StatusMessageService)), logger)
+        } catch (error) {
+          // If we come in here, we had an error thrown which means we should fail the test.
+          expect(true).toBeFalsy()
```
KvelaGorrrrnio: Why not just remove the try-catch and let the test runner catch the error?

src/business-logic/services/jsend-http-service.ts
```diff
@@ -0,0 +1,52 @@
...
+interface JsendErrorResponse {
+  status: RequestStatus.ERROR
+  message: string
+  code: number
+}
+
+interface JsendValidResponse {
+  status: RequestStatus.SUCCESS | RequestStatus.FAIL
+  data: unknown
+}
```
KvelaGorrrrnio: I know that the data attribute is on both, but I don't think that they should be a single interface. It also raises the question of what a valid response is. If these two are a valid response, then I would argue that `JsendErrorResponse` is as well.
If we don't use fail at all or intend to, then It can also be removed for now and be introduced if needed.

src/business-logic/services/jsend-http-service.ts
```diff
@@ -0,0 +1,52 @@
...
+      case RequestStatus.ERROR: {
+        throw new HttpException(response.code, response.message)
+      }
+      default: {
+        throw new HttpException(HttpStatusCode.INTERNAL_SERVER_ERROR, 'Couldn\'t recognize Jsend Response')
+      }
+    }
+  }
+
+  private validate(data: unknown): asserts data is JsendResponse {
```
KvelaGorrrrnio: Perhaps this should be renamed to something like `assertJsendResponse`?

src/business-logic/services/jsend-http-service.ts
```diff
@@ -0,0 +1,52 @@
...
+  status: RequestStatus.SUCCESS | RequestStatus.FAIL
+  data: unknown
+}
+
+export class JsendHttpService implements HttpService {
+  public constructor(private readonly httpService: HttpService) {
+  }
+
+  public async get(url: string): Promise<unknown> {
+    const response = await this.httpService.get(url)
```
KvelaGorrrrnio: What is the type of the response? (I assume unknown)

src/business-logic/services/jsend-http-service.ts
```diff
@@ -0,0 +1,52 @@
...
+    }
+  }
+
+  private validate(data: unknown): asserts data is JsendResponse {
+    if (typeof data !== 'object' || data === null) {
+      throw new UnsupportedOperationException('Expected object')
+    }
+
+    if (!('status' in data)) {
+      throw new UnsupportedOperationException('Response had not attribute.adsfasf')
```
KvelaGorrrrnio: This is a very good error message :+1:
One could consider using something like:
```typescript
throw new UnsupportedOperationException(`Expected JSend response to have the attribute 'status'.`)
```

src/business-logic/services/jsend-http-service.ts
```diff
@@ -0,0 +1,52 @@
...
+      }
+      default: {
+        throw new HttpException(HttpStatusCode.INTERNAL_SERVER_ERROR, 'Couldn\'t recognize Jsend Response')
+      }
+    }
+  }
+
+  private validate(data: unknown): asserts data is JsendResponse {
+    if (typeof data !== 'object' || data === null) {
+      throw new UnsupportedOperationException('Expected object')
```
KvelaGorrrrnio: A bit more information would be nice, e.g.
```typescript
throw new UnsupportedOperationException(`Expected JSend response to be an object, but received `${typeof data}`.`)
```

src/business-logic/services/status-message-service.ts
```diff
@@ -0,0 +1,26 @@
...
+export class StatusMessageService {
+  public constructor(
+    private readonly statusMessageRepository: StatusMessageRepository,
+    private readonly statusMessageEventEmitter: StatusMessageEventEmitter,
+    private readonly httpService: HttpService
+  ) {
+  }
+
+  public async getStatusMessages(): Promise<StatusMessage[]> {
+    const statusMessageFromAlbaServer: StatusMessage[] = await this.httpService.get(STATUS_MESSAGE_URL) as StatusMessage[]
```
KvelaGorrrrnio, line 16: I like how the `httpService` forces us to make it explicit when we assume the shape of external data :+1: It makes it easy to determine where we might want to validate or parse data into a shape.

src/model/entities/status-message.ts
KvelaGorrrrnio: I would still like to see this as a required attribute :D

src/presentation/fastify-server.ts
```diff
@@ -37,10 +45,26 @@ export class FastifyServer implements ProxyServer {
     })
   }
 
+  // The below eslint-disable is necessary.
+  // If we add 'await' or .then().catch() we will get a 'MaxListenerExceededWarning: Possible EventEmitter memory leak detected.'
+  /* eslint-disable @typescript-eslint/no-floating-promises */
+  private addCors(): void {
+    this.fastifyServer.addHook('onRequest', async(request, reply) => {
```
KvelaGorrrrnio, line 49: It would be nice with a space between `async` and `(`.

src/presentation/fastify-server.ts
```diff
@@ -37,10 +45,26 @@ export class FastifyServer implements ProxyServer {
...
+      reply.header('Access-Control-Allow-Origin', '*')
+      reply.header('Access-Control-Allow-Credentials', true)
+      reply.header('Access-Control-Allow-Headers', 'Authorization, Origin, X-Requested-With, Content-Type, Accept, X-Slug, X-UID')
+      reply.header('Access-Control-Allow-Methods', 'OPTIONS, POST, PUT, PATCH, GET, DELETE')
+      if (request.method === 'OPTIONS') {
+        reply.send()
+      }
+    })
+  }
```
KvelaGorrrrnio: The following suggestion works and doesn't need eslint-disable:
```typescript
  private addCors(): void {
    this.fastifyServer.addHook('onRequest', async(request, reply) => {
      reply
        .header('Access-Control-Allow-Origin', '*')
        .header('Access-Control-Allow-Credentials', true)
        .header('Access-Control-Allow-Headers', 'Authorization, Origin, X-Requested-With, Content-Type, Accept, X-Slug, X-UID')
        .header('Access-Control-Allow-Methods', 'OPTIONS, POST, PUT, PATCH, GET, DELETE').then(() => {}, () => {})
      if (request.method === 'OPTIONS') {
        reply.send().then(() => {}, () => {})
      }
    })
  }
```

KvelaGorrrrnio (approved):

src/business-logic/services/jsend-http-service.ts
RasmusAlbrektsen: I think the original is fine, but otherwise we should probably use the suggestion here.

src/business-logic/services/status-message-service.ts
```diff
@@ -0,0 +1,26 @@
...
+
+const STATUS_MESSAGE_URL: string = '/systemInformation/statusMessages'
+
+export class StatusMessageService {
+  public constructor(
+    private readonly statusMessageRepository: StatusMessageRepository,
+    private readonly statusMessageEventEmitter: StatusMessageEventEmitter,
+    private readonly httpService: HttpService
+  ) {
+  }
```
RasmusAlbrektsen: Personal preference: I like it when we close the bracket on line 13 here since it is empty anyway.

src/presentation/services/status-message-event-service.ts
```diff
@@ -0,0 +1,39 @@
...
+    if (!this.instance) {
+      this.instance = new StatusMessageEventService()
+    }
+    return this.instance
+  }
+
+  private readonly callbacks: ((statusMessageEvent: StatusMessageEvent) => void)[] = []
+
+  private constructor() {
+  }
```
RasmusAlbrektsen: I'd move this closing bracket to line 19 to be consistent, also to be in line with my earlier comment.

RasmusAlbrektsen (approved):

src/business-logic/panel-integrations/skaarhoj-panel.ts
```diff
@@ -0,0 +1,75 @@
...
+      this.socket.write('list\n')
+      this.statusMessageService.sendStatusMessage(this.createConnectionSuccessStatusMessage())
+    })
+
+    this.socket.on('error', (error) => {
+      this.logger.data(error).error(`Error from ${SkaarhojPanel.name}:${this.panelConfiguration.hostname}`)
+      this.statusMessageService.sendStatusMessage(this.createUnableToConnectStatusMessage())
+    })
+
+    this.socket.setEncoding('utf8')
```
MadsThy, line 51: Is there a reason the utf8 encoding is not set at the beginning of this method? :)

src/business-logic/services/jsend-http-service.ts
MadsThy: Im personally not fuzzy, but i think OIJSDFIOJSD beats asdfasf any day of the week ;)

src/presentation/services/http-error-handler.ts
```diff
@@ -15,27 +19,30 @@ export class HttpErrorHandler {
...
+    if (exception instanceof NotFoundException) {
+      return HttpStatusCode.NOT_FOUND
+    }
+    if (exception instanceof UnsupportedOperationException) {
+      return HttpStatusCode.BAD_REQUEST
+    }
+    if (exception instanceof DatabaseNotConnectedException) {
+      return HttpStatusCode.SERVICE_UNAVAILABLE
+    }
+    return HttpStatusCode.INTERNAL_SERVER_ERROR
```
MadsThy, line 46: Is the norm not to use a switch case when having 3+ options? :)

MadsThy (approved):

src/presentation/fastify-server.ts
LindvedKrvang: I would like it too. It is the lint-rule, that ensures that there is no space between a function name and its parantheses, that does this :p

src/data-access/in-memory/in-memory-status-message-repository.ts
LindvedKrvang: This constructor isn't useless. In fact, it serves quite an important purpose: It forces users of the class to go through the `getInstance` method instead of directly instantiating a new instance.

src/business-logic/panel-integrations/skaarhoj-panel.ts
LindvedKrvang: I assume no. I just copied the code from the `skaarhoj-connection-demo` :p

src/presentation/services/http-error-handler.ts
LindvedKrvang: It's not, but I couldn't get `instanceof` to work with the switch. Suggestions are appreciated

src/business-logic/services/jsend-http-service.ts
LindvedKrvang: Seems like we need a good old fashion street-fight to settle this

src/business-logic/services/jsend-http-service.ts
LindvedKrvang: Not even the compiler knows ;)

src/business-logic/panel-integrations/skaarhoj-panel.ts
LindvedKrvang: updated

src/business-logic/panel-integrations/skaarhoj-panel.ts

src/business-logic/panel-integrations/skaarhoj-panel.ts

src/business-logic/panel-integrations/test/skaarhoj-panel.spec.ts
LindvedKrvang: Updated

src/business-logic/services/jsend-http-service.ts
LindvedKrvang: Removing FAIL for now

src/business-logic/services/jsend-http-service.ts
LindvedKrvang: Updated

src/business-logic/services/jsend-http-service.ts
LindvedKrvang: Updated

src/model/entities/status-message.ts
LindvedKrvang: Updated

src/business-logic/services/status-message-service.ts
LindvedKrvang: Same preference here. It's been updated.

src/presentation/services/status-message-event-service.ts
LindvedKrvang: Updated

src/model/entities/status-message.ts
LindvedKrvang: Updated

src/presentation/fastify-server.ts
LindvedKrvang: Nice :) 
It's been updated

src/presentation/services/http-error-handler.ts
MadsThy:
```suggestion
    private mapExceptionToHttpStatusCode(exception: Exception): HttpStatusCode {
    switch (true) {
      case exception instanceof HttpException:
        return (exception as HttpException).httpStatusCode
      case exception instanceof NotFoundException:
        return HttpStatusCode.NOT_FOUND
      case exception instanceof UnsupportedOperationException:
        return HttpStatusCode.BAD_REQUEST
      case exception instanceof DatabaseNotConnectedException:
        return HttpStatusCode.SERVICE_UNAVAILABLE
      default:
        return HttpStatusCode.INTERNAL_SERVER_ERROR
    }
  }
```

How about this? :)

MadsThy (approved):

src/presentation/fastify-server.ts
KvelaGorrrrnio: Ohhh. I thought I had updated it for async lambdas. I guess that is a change for another PR then :)

KvelaGorrrrnio (approved):

src/presentation/services/http-error-handler.ts
LindvedKrvang: There is something abouth the `switch (true)` that doesn't sit right with me...

PR #39, opened 2024-10-28
Sof 2248/connect to panel on panel creation

LindvedKrvang:
Note: Dependent on PR https://github.com/tv2/sofie-tv2-server/pull/38

When creating a new PanelConfiguration, we now try and connect to immediatly.

If there is already a Panel for the hostname of the PanelConfiguration, then we don't try and connect to it, so we don't override the established connection.


MadsThy (approved):

src/business-logic/services/panel-manager-implementation.ts
```diff
@@ -26,9 +42,18 @@ export class PanelManagerImplementation implements PanelManager {
...
     const panelConfigurations: PanelConfiguration[] = await this.panelConfigurationRepository.getPanelConfigurations()
-    panelConfigurations.map((panelConfiguration) => {
-      const panel: Panel = this.panelFactory.createPanel(panelConfiguration)
-      panel.initialize()
-    })
+    panelConfigurations.map(this.connectToPanel.bind(this))
+  }
+
+  private connectToPanel(panelConfiguration: PanelConfiguration): void {
+    if (this.panels.has(panelConfiguration.hostname)) {
```
RasmusAlbrektsen, line 49: I'm a bit confused about the use case for this scenario to be necessary. Shouldn't we from the client handle it beforehand. If we set up locally simulated panels, i'm guessing we would have an issue getting through this too? (or if they are exposed by a proxy perhaps).

src/business-logic/services/panel-manager-implementation.ts
```diff
@@ -4,16 +4,32 @@ import { PanelConfigurationRepository } from '../../data-access/interfaces/panel
 import { PanelConfiguration } from '../../model/interfaces/panel-configuration'
 import { PanelFactory } from '../panel-integrations/panel-factory'
 import { Panel } from './interfaces/panel'
+import { PanelEventObserver } from '../../presentation/interfaces/panel-event-observer'
+import { PanelConfigurationCreatedEvent, PanelEvent } from '../../presentation/value-objects/panel-event'
+import { PanelEventType } from '../../presentation/enums/event-type'
 
 export class PanelManagerImplementation implements PanelManager {
   private readonly logger: Logger
```
RasmusAlbrektsen, line 13: I wouldn't have this newline between equally accessible variables. Is there any reason you prefer this, if intentional? :-)

RasmusAlbrektsen (approved):

src/business-logic/services/panel-manager-implementation.ts
LindvedKrvang: That is asuming that there will only ever be one client. 
It's a good idea to not trust blindly the data the backend receives from clients.

Anyways: We need some way of handling two Panels connecting to the same hostname. I'm up for ideas. This was just to have something :)

src/business-logic/services/panel-manager-implementation.ts
LindvedKrvang: Not sure what you are referring to. There is only one variable? The newline is between the variable and the constructor (which I would like to keep).

I would however, like to have a newline between the class declaration and the first variable.

PR #40, opened 2024-10-28
SOF-2250 Disconnect from Skaarhoj panel when deleting a PanelConfiguration

LindvedKrvang:
Note: Dependent on https://github.com/tv2/sofie-tv2-server/pull/39

When deleting a `PanelConfiguration` we now also disconnected from the Panel if there is a Panel connected for it.


RasmusAlbrektsen (approved):

src/business-logic/services/panel-manager-implementation.ts
```diff
@@ -24,6 +28,11 @@ export class PanelManagerImplementation implements PanelManager {
     this.panelEventObserver.subscribeToPanelEvents((panelEvent: PanelEvent) => {
       if (this.isPanelConfigurationCreatedEvent(panelEvent)) {
         this.connectToPanel(panelEvent.panelConfiguration)
+        return
+      }
+      if (this.isPanelConfigurationDeletedEvent(panelEvent)) {
+        this.disconnectFromPanel(panelEvent.panelConfigurationId)
+        return
```
MadsThy, line 35: is this return really needed? 😄

MadsThy (approved):

src/business-logic/services/panel-manager-implementation.ts
LindvedKrvang: For the current implementation, it's not

PR #41, opened 2024-10-30
SOF-2249 Reconnect to Panel on PanelConfiguration updates

LindvedKrvang:
When a PanelConfiguration is updated, we disconnect the current Panel for it, delete the Panel and create a new Panel from the updated PanelConfiguration.

The reason for disconnecting and discarding is, that the new PanelConfiguration could be incompatible with the current Panel (maybe it changed from Skaarhoj to StreamDeck - not currently supported, but it showcases the usecase).


src/business-logic/services/panel-manager-implementation.ts
```diff
@@ -34,6 +64,10 @@ export class PanelManagerImplementation implements PanelManager {
         this.disconnectFromPanel(panelEvent.panelConfigurationId)
         return
       }
+      if (this.isPanelConfigurationUpdatedEvent(panelEvent)) {
+        this.reconnectToPanel(panelEvent.panelConfiguration)
+        return
```
MadsThy, line 69: im pretty sure you don't need this return 😄 
```suggestion
```

MadsThy (approved):

src/business-logic/services/panel-manager-implementation.ts
LindvedKrvang: But now the return from the delete was necessary ;)

LindvedKrvang: This is going to get an overhaul in SOF-2261.
In my current implementation I'm breaking the rules/barriers of our architecture which is not allowed... :p

src/business-logic/services/panel-manager-implementation.ts
```diff
@@ -24,7 +25,36 @@ export class PanelManagerImplementation implements PanelManager {
...
+    try {
+      await this.connectToPanels()
+    } catch (error) {
+      this.logger.data(error).error('Error while connecting to Panels')
+    }
+  }
+
+  private async connectToPanels(): Promise<void> {
+    const panelConfigurations: PanelConfiguration[] = await this.panelConfigurationRepository.getPanelConfigurations()
+    panelConfigurations.map(this.connectToPanel.bind(this))
```
KvelaGorrrrnio: `forEach` can be used here, since nothing is returned.

KvelaGorrrrnio (approved):

src/business-logic/services/panel-manager-implementation.ts
LindvedKrvang: Updated

PR #42, opened 2024-10-30
SOF-2253 Reconnect on connection lost

LindvedKrvang:
Note: Dependent on https://github.com/tv2/sofie-tv2-server/pull/41

Reconnect to a Skaarhoj Panel when we lose connection that wasn't on purpose


MadsThy (approved):

src/business-logic/panel-integrations/skaarhoj-panel.ts
```diff
@@ -95,9 +102,38 @@ export class SkaarhojPanel implements Panel {
...
+
+      this.connectToSocket()
+    }, RECONNECTION_TIMEOUT_MS)
+  }
+
+  private createReconnectingStatusMessage(): StatusMessage {
+    return {
+      id: this.getStatusMessageId(),
+      title: 'Reconnecting to Skaarhoj panel',
+      message: `Attempting to reconnect to the Skaarhoj panel at ${this.panelConfiguration.hostname}`,
```
KvelaGorrrrnio, line 131: It would be nice with a `.` at the end of the sentence. It helps to determine that the information given ends here and confirm that something isn't missing :)

src/business-logic/panel-integrations/skaarhoj-panel.ts
```diff
@@ -9,11 +9,15 @@ import { StatusMessage } from '../../model/entities/status-message'
...
 
 const SKAARHOJ_PORT: number = 9923
+const RECONNECTION_TIMEOUT_MS: number = 5000
 
 export class SkaarhojPanel implements Panel {
   private readonly logger: Logger
   private socket: Socket = new Socket()
 
+  private keepAlive: boolean = true
+  private reconnectionTimeout: NodeJS.Timeout | undefined
```
KvelaGorrrrnio: It might just be me, but I found the `ion` part a bit odd. I would just write `reconnectTimeout` :D To make it a bit clearer, one could add a `Identifier` suffix, so it becomes: `reconnectTimeoutIdentifier?: NodeJS.Timeout`

KvelaGorrrrnio (approved):

RasmusAlbrektsen (approved):

src/business-logic/panel-integrations/skaarhoj-panel.ts
LindvedKrvang: Updated.
Had it been a log statement, then I would have disagreed.

src/business-logic/panel-integrations/skaarhoj-panel.ts
LindvedKrvang: Updated to `reconnectTimeoutIdentifier`

PR #44, opened 2024-10-31
Sof 2261/map Panel inputs to call actions in Alba

LindvedKrvang:
Note: Dependent on https://github.com/tv2/sofie-tv2-server/pull/43

When an input on a Skaarhoj Panel is registered, we now find the corresponding `InputConfiguration` for the input and map it to a `PanelCommand`. This `PanelCommand` is given back to the `PanelManager` which hands it over to the `PanelCommandExecutor` to execute the Command. In the current implementation we handle Commands that calls an "Action" in Alba. (The quotation marks is because we also have "pseudo" Actions. A "pseudo" Action is not Action that have a corresponding Action entity, but a functionality to be called in Alba i.e. Take or SetNext.)

Also in this PR:
- I broke our architecture rules in some earlier PRs. This is fixed by a refactor to introduce an `EventBus` which allows us to listen for events within our logic layer without being dependent on Presentation specific code (for example, before we were passing DTOs around in the logic layer...). The EventBus logic is heavily inspired by the EventBus implemented in the new INews-gateway.
- Refactor `InputData` to be `PanelCommand`.
- Refactor the `HttpMethodInputData` to be `ActionPanelCommand`.


src/business-logic/facades/domain-event-facade.ts
```diff
@@ -0,0 +1,30 @@
...
+  }
+
+  public static createStatusMessageEmitter(): StatusMessageEmitter {
+    return this.eventBus
+  }
+
+  private static getEventBus(): EventBus {
+    this.eventBus ??= new EventBus()
+    return this.eventBus
+  }
```
KvelaGorrrrnio, line 29: Me like 👏

src/business-logic/facades/domain-event-facade.ts
```diff
@@ -0,0 +1,30 @@
...
+  public static createPanelObserver(): PanelObserver {
+    return this.getEventBus()
+  }
+
+  public static createPanelEmitter(): PanelEmitter {
+    return this.getEventBus()
+  }
+
+  public static createStatusMessageObserver(): StatusMessageObserver {
+    return this.eventBus
```
MadsThy, line 19: So... I see 4 methods, and they all basically do the same, but... not... the same. What's going on here? 😄

src/business-logic/interfaces/panel.ts
```diff
@@ -0,0 +1,11 @@
+import { PanelConfiguration } from '../../model/interfaces/panel-configuration'
+import { PanelLayoutConfiguration } from '../../model/interfaces/panel-layout-configuration'
+import { PanelCommand } from '../../model/interfaces/input-configuration'
+
+export interface Panel {
+  initialize(): void
```
KvelaGorrrrnio: Is initialize intended to have separate functionality than preparing for and connecting to a panel?
If not, then one could consider renaming it to `connect`, to give an indication of that it is what we expect happens when we call it.

src/business-logic/panel-integrations/skaarhoj-panel.ts
```diff
@@ -145,4 +176,59 @@ export class SkaarhojPanel implements Panel {
...
+    const match = commandArray[commandArray.length - 1]?.match(SKAARHOJ_INPUT_REGEX)
+    if (!match || !match.groups) {
+      return
+    }
+
+    if (match.groups.inputType !== SKAARHOJ_INPUT_PREFIX) {
+      return
+    }
+
+    // It's possible to get an id like "3.4", so the Math.floor is to turn that into "3" for now since we don't support multiple functions for a single button.
```
KvelaGorrrrnio: If I was a new developer looking at this, I would wonder why one could get 3.4. Perhaps it should state why or state what is meant with multiple functions. E.g.
```typescript
// It is possible to get an id like "3.4", which indicates a specific area of button 3 is pressed or released.
// Since we treat the areas on a button as one, `Math.floor` is used to get the button identifier. For the id "3.4" it would be "3".

File: src/business-logic/panel-integrations/skaarhoj-panel.ts [hunk 5]
```diff
@@ -145,4 +176,59 @@ export class SkaarhojPanel implements Panel {
...
+    const id: number = Math.floor(Number.parseFloat(match.groups.id!))
```
KvelaGorrrrnio: One could also use a replace to keep it as a string:
```typescript
const id: string = match.groups.id!.replace(/\..*$/, '')
```
or do a split on '.':
```typescript
const id: string = match.groups.id!.split('.')[0]

File: src/business-logic/panel-integrations/skaarhoj-panel.ts [hunk 6]
```diff
@@ -145,4 +176,59 @@ export class SkaarhojPanel implements Panel {
...
+    const inputConfiguration: InputConfiguration | undefined = this.panelLayoutConfiguration.inputConfigurations[id]
+    if (!inputConfiguration) {
+      return
+    }
+
+    switch (inputConfiguration.type) {
+      case InputType.BUTTON: {
+        const isButtonPressed: boolean = !!input.toUpperCase().match(SkaarhojInputType.BUTTON_PRESSED) && !!inputConfiguration.onPress
+        const isButtonReleased: boolean = !!input.toUpperCase().match(SkaarhojInputType.BUTTON_RELEASED) && !!inputConfiguration.onRelease
```
KvelaGorrrrnio: The value of these variables reflects more than what the name states. They check whether a button is pressed or released and if the button configuration is configured for press or release.

The `ButtonConfiguration` attributes have both `onPress` and `onRelease`. Do we need to be able to support triggering on both events?
If not, then I would suggest that we do something similar to what is done in the frontend with `KeyEventType`, which has the values PRESSED and RELEASED. Then a buttonConfiguration could have `triggersOn: KeyEventType`. If that is done, then the logic for this could also be simplified to:

1. Get the `KeyEventType` of the configuration.
2. Parse the `KeyEventType` from the input string (it might be undefined if it was not found).
3. Compare equality between them to determine if the command should be returned.

src/business-logic/panel-integrations/skaarhoj-panel.ts
```diff
@@ -1,27 +1,55 @@
...
+ * - HWC#1=UP
+ * - HWC#21=DOWN
+ * - HWC#50=Abs:50
+ *
+ * The 'HWC' is a prefix given to all Skaarhoj inputs. That's the 'inputType' group of the regex.
+ * The number after the '#' is the id of the input i.e., which id it is. That's the 'id' group of the regex.
+ * The right side of the equals sign corresponds to what happened to the input. It could be that the input is a button that was pressed or released.
+ * It could also be that the input is a fader, and it tells us what is the new numeric value of the fader. That's the 'data' group of the regex.
+ */
+const SKAARHOJ_INPUT_REGEX: RegExp = /(?<inputType>[a-z]+)#(?<id>(\d+)|(\d+.\d+))=(?<data>.*)/i
```
RasmusAlbrektsen, line 28: I actually really appreciate the explanation in the comment above!

src/business-logic/panel-integrations/skaarhoj-panel.ts
```diff
@@ -145,4 +176,59 @@ export class SkaarhojPanel implements Panel {
...
+    // It's possible for Skaarhoj to send an array of commands separated by '\n'. We want the last entry of that array
+    const commandArray: string[] = input.split('\n').filter(s => s !== '')
+    const match = commandArray[commandArray.length - 1]?.match(SKAARHOJ_INPUT_REGEX)
+    if (!match || !match.groups) {
+      return
+    }
+
+    if (match.groups.inputType !== SKAARHOJ_INPUT_PREFIX) {
+      return
+    }
```
KvelaGorrrrnio: This method has 2 responsibilities. The first being to parse the input string and get the last one. The second being doing the mapping.
I will suggest to move the parsing out into a separate method and perhaps change the mapping method to take a structured input, like so:

```typescript
interface SkaarhojInput {
  inputType: string
  id: string
  data: string
}

class SkaarhojPanel {
  private parseSkaarhojInput(text: string): SkaarhojInput | undefined {
    // It's possible for Skaarhoj to send an array of commands separated by '\n'. We want the last entry of that array
    const inputs: string[] = text.split('\n').filter(s => s !== '')
    const input: SkaarhojInput | undefined = commandArray[commandArray.length - 1]?.match(SKAARHOJ_INPUT_REGEX)?.groups as SkaarhojInput | undefined
    if (!input) {
      return
    }

    if (input.inputType !== SKAARHOJ_INPUT_PREFIX) {
      return
    }

    // It's possible to get an id like "3.4", so the Math.floor is to turn that into "3" for now since we don't support multiple functions for a single button.
    const id: number = Math.floor(Number.parseFloat(input.id))

    return {
      ...input,
      id: id.toString(),
    }
  }

  private mapPanelInputToCommand(input: SkaarhojInput): PanelCommand | undefined {
    const inputConfiguration: InputConfiguration | undefined = this.panelLayoutConfiguration.inputConfigurations[input.id]
    if (!inputConfiguration) {
      return
    }

    switch (inputConfiguration.type) {
      case InputType.BUTTON: {
        const isButtonPressed: boolean = !!input.data.toUpperCase().match(SkaarhojInputType.BUTTON_PRESSED) && !!inputConfiguration.onPress
        const isButtonReleased: boolean = !!input.data.toUpperCase().match(SkaarhojInputType.BUTTON_RELEASED) && !!inputConfiguration.onRelease

        if (!isButtonPressed && !isButtonReleased) {
          return
        }
        return inputConfiguration.command
      }
      case InputType.FADER: {
        if (!input.data.toUpperCase().match(SkaarhojInputType.FADER)) {
          return
        }
        const regexValue = input.data.match(/Abs:(?<value>\d+)/)
        if (!regexValue) {
          return
        }
        const value: number = Number.parseFloat(regexValue[0].replace('Abs:', ''))
        if (inputConfiguration.command.type !== PanelCommandType.T_BAR) {
          return
        }
        inputConfiguration.command.value = value
        return inputConfiguration.command
      }
    }
  }
}
```

src/business-logic/panel-integrations/skaarhoj-panel.ts
```diff
@@ -145,4 +176,59 @@ export class SkaarhojPanel implements Panel {
...
+  }
+
+  public registerOnCommand(onCommandCallback: (command: PanelCommand) => void): void {
+    this.onCommandCallback = onCommandCallback
+  }
+
+  private mapPanelInputToCommand(input: string): PanelCommand | undefined {
+    // It's possible for Skaarhoj to send an array of commands separated by '\n'. We want the last entry of that array
+    const commandArray: string[] = input.split('\n').filter(s => s !== '')
+    const match = commandArray[commandArray.length - 1]?.match(SKAARHOJ_INPUT_REGEX)
```
KvelaGorrrrnio: One could also use `Array.findLast`, so it becomes:
```typescript
const lastInput: SkaarhojInput | undefined = input.split('\n')
  .findLast(text => text.trim() !== '')
  ?.match(SKAARHOJ_INPUT_REGEX)?.groups as SkaarhojInput | undefined
```

src/business-logic/panel-integrations/skaarhoj-panel.ts
```diff
@@ -145,4 +176,59 @@ export class SkaarhojPanel implements Panel {
...
+    if (!match || !match.groups) {
+      return
+    }
```
KvelaGorrrrnio: Optional chaining can be used to simplify the condition to `!match?.groups`.

src/business-logic/services/alba-server-http-service.ts
```diff
@@ -7,9 +7,17 @@ export class AlbaServerHttpService implements HttpService {
...
+  }
+
+  private prefixUrlWithSlash(url: string): string {
     if (!url.startsWith('/')) {
-      url = `/${url}`
+      return `/${url}`
     }
-    return this.httpService.get(`${ALBA_API_BASE_URL}${url}`)
+    return url
+  }
```
KvelaGorrrrnio: Regex suggestion: `return url.replace(/^\/*/, '/')`
It matches zero or more forward slashes at the beginning of the string, so it also removes double slashes.

src/business-logic/interfaces/panel.ts
```diff
@@ -0,0 +1,11 @@
...
+  disconnect(): void
+  getPanelConfiguration(): PanelConfiguration
```
KvelaGorrrrnio, line 8: Wouldn't it suffice to only get the id?

src/business-logic/interfaces/http-service.ts
```diff
@@ -1,3 +1,4 @@
 export interface HttpService {
   get(url: string): Promise<unknown>
+  put(url: string, body?: unknown): Promise<void>
```
MadsThy, line 3: should body really be optional? :)

src/business-logic/panel-integrations/skaarhoj-panel.ts
```diff
@@ -57,7 +85,10 @@ export class SkaarhojPanel implements Panel {
 
     this.socket.setEncoding('utf8')
     this.socket.on('data', (data) => {
-      this.logger.data(data).info(`Received input from ${SkaarhojPanel.name}:${this.panelConfiguration.hostname}`)
+      const inputData: PanelCommand | undefined = this.mapPanelInputToCommand(data.toString())
```
MadsThy:
```suggestion
      const inputData: PanelCommand | undefined = this.mapPanelInputToCommand(`${data}`)
```

src/business-logic/panel-integrations/skaarhoj-panel.ts
```diff
@@ -145,4 +176,59 @@ export class SkaarhojPanel implements Panel {
...
+
+  public registerOnCommand(onCommandCallback: (command: PanelCommand) => void): void {
+    this.onCommandCallback = onCommandCallback
+  }
+
+  private mapPanelInputToCommand(input: string): PanelCommand | undefined {
+    // It's possible for Skaarhoj to send an array of commands separated by '\n'. We want the last entry of that array
+    const commandArray: string[] = input.split('\n').filter(s => s !== '')
+    const match = commandArray[commandArray.length - 1]?.match(SKAARHOJ_INPUT_REGEX)
+    if (!match || !match.groups) {
```
MadsThy:
```suggestion
    if (!match?.groups) {
```
Would this work? :)

src/business-logic/panel-integrations/skaarhoj-panel.ts
```diff
@@ -145,4 +176,59 @@ export class SkaarhojPanel implements Panel {
...
+        if (!isButtonPressed && !isButtonReleased) {
+          return
+        }
+        return inputConfiguration.command
+      }
+      case InputType.FADER: {
+        if (!input.toUpperCase().match(SkaarhojInputType.FADER)) {
+          return
+        }
+        const regexValue = match.groups.data?.match(/Abs:(?<value>\d+)/)
```
MadsThy: Could it be an idea to name this const what you are looking for here? :)

src/business-logic/services/alba-server-http-service.ts
MadsThy: Suggestion:

```suggestion
private prefixUrlWithSlash(url: string): string {
  return url.startsWith('/') ? url : `/${url}`;
}
```

src/business-logic/services/alba-server-http-service.ts
```diff
@@ -7,9 +7,17 @@ export class AlbaServerHttpService implements HttpService {
...
+
+  public put(url: string, body?: unknown): Promise<void> {
```
MadsThy, line 17: I dont think the body should be optional 😄

src/business-logic/services/panel-command-executor.ts
```diff
@@ -0,0 +1,69 @@
...
+  SET_NEXT_SEGMENT = 'SET_NEXT_SEGMENT',
+  SET_PREVIOUS_SEGMENT = 'SET_PREVIOUS_SEGMENT'
+  // TODO: Add missing "pseudo" Actions as needed
+}
+
+export class PanelCommandExecutor {
+  private readonly logger: Logger
+
+  public constructor(private readonly httpService: HttpService, logger: Logger) {
+    this.logger = logger.tag(PanelCommandExecutor.name)
```
MadsThy, line 22:
```suggestion
    this.logger = logger.tag(this.constructor.name)
```

MadsThy (approved):

KvelaGorrrrnio (approved):
Looks good :) The only thing that caught my eye is the `onPress`/`onRelease` attributes for `ButtonConfiguration`.

src/business-logic/facades/domain-event-facade.ts
LindvedKrvang: It's called Interface Segregation. The I of SOLID.

src/business-logic/interfaces/http-service.ts
LindvedKrvang: Maybe not.
But we have calls to PUT requests in the backend that doesn't expect a body.
It falls back to the underlining issue: How do we handle request, that does not return data, but instead "execute something" in the backend, i.e. telling the backend to do a Take etc?

The current way we are doing this is by using PUT request.

If `body` wasn't optional; What would a consumer of `put()` give as a body, when there is no body needed?

src/business-logic/services/alba-server-http-service.ts
LindvedKrvang: Refer to my other comment

src/business-logic/interfaces/panel.ts
LindvedKrvang: In the current implementation, no. I have a usecase where I need the `hostname` of the PanelConfiguration.

src/business-logic/services/panel-command-executor.ts
LindvedKrvang: What's the difference?

src/business-logic/panel-integrations/skaarhoj-panel.ts
LindvedKrvang: You are not a fan of the `.toString()` method?

src/business-logic/panel-integrations/skaarhoj-panel.ts
LindvedKrvang: It would and it has been updated :)

src/business-logic/panel-integrations/skaarhoj-panel.ts
LindvedKrvang: Yes.
I had a bit of trouble of coming up with a good name in the first place, hence the name `regexValue`.
I have updated it to `regexMatchForFaderValue` and added its types.

src/business-logic/services/alba-server-http-service.ts
LindvedKrvang: I was going to implement your suggestion, but then I saw Anders suggestion and went with that one... :p

src/business-logic/interfaces/panel.ts
LindvedKrvang: Good suggestion. It's been updated.

src/business-logic/panel-integrations/skaarhoj-panel.ts
LindvedKrvang: Updated to include the "why". Modified the comment to match the new way of stripping the decimals.

src/business-logic/panel-integrations/skaarhoj-panel.ts
LindvedKrvang: Updated to your first suggestion.

src/business-logic/panel-integrations/skaarhoj-panel.ts
LindvedKrvang: Updated.
This code snippet was originally written in a repo where the Node version did not support `findLast()`, so it's nice to get to use it :)

src/business-logic/panel-integrations/skaarhoj-panel.ts

src/business-logic/services/alba-server-http-service.ts
LindvedKrvang: Nice. It's been updated.

src/business-logic/panel-integrations/skaarhoj-panel.ts
LindvedKrvang: The responsibilities has been split

src/business-logic/panel-integrations/skaarhoj-panel.ts
LindvedKrvang: As far as I know then no, we don't need to support triggering on both events.
I have updated it to your suggestion.
`onPress` and `onRelease` has been replaced with `triggersOn` on `ButtonInputConfiguration`.

RasmusAlbrektsen (approved):
Great work :-)

src/business-logic/services/panel-command-executor.ts
KvelaGorrrrnio: They are the same functionality-wise, but if we e.g. rename the class or copy the file and changes the name, then `this.constructor.name` will reflect the correct name, no matter if the IDE has excellent refactoring capabilities or not.

KvelaGorrrrnio (approved):

src/business-logic/services/panel-command-executor.ts
LindvedKrvang: But if you change the name of the class without changing this, then you get a compilation error, so the name change will be reflected.

PR #46, opened 2024-11-01
Sof 2262/implement modifiers

LindvedKrvang:
Note: Is dependent on https://github.com/tv2/sofie-tv2-server/pull/44

We now support having keys that requires "modifiers" to be activated.

For now I have introduced the `PanelInputModifier` (which has four options for modifiers). In theory, we could just let it be a random string, so Clients would be able to define however many modifiers they want. I opted to have a "type" for it for now. It's only used in validation of the incoming data.

I have also opted to have `;` as a delimiter, so we have some way of tracking modifiers and the actual key within the key for an InputConfiguration. So keys can look like: `{key}`, `MODIFIER_ONE;{key}` or `MODIFIER_TWO;MODIFIER_FOUR;{key}`.

Suggestions are as always most welcome :) 

Also in this PR:
- `Panel` is now an abstract class instead of interface so we don't need to reimplement common Panel behaviour in each Panel integration (for example, modifier behaviour)
- "Panel Groups" has been introduced. The `PanelConfiguration` now takes a `panelGroupId` and the PanelManager will keep track of the groups. Modifiers are only active within its own PanelGroup.


src/business-logic/panel-integrations/skaarhoj-panel.ts
```diff
@@ -199,7 +185,7 @@ export class SkaarhojPanel implements Panel {
 
     // It's possible to get an id like "3.4", so the Math.floor is to turn that into "3" for now since we don't support multiple functions for a single button.
     const id: number = Math.floor(Number.parseFloat(match.groups.id!))
-    const inputConfiguration: InputConfiguration | undefined = this.panelLayoutConfiguration.inputConfigurations[id]
+    const inputConfiguration: InputConfiguration | undefined = this.getInputConfiguration(id + '')
```
MadsThy: if you change line 187 to 
```typescript
const id: string = `${Math.floor(Number.parseFloat(match.groups.id!))}`
```

```suggestion
    const inputConfiguration: InputConfiguration | undefined = this.getInputConfiguration(id)
```

Or you could do this :D 

```suggestion
    const inputConfiguration: InputConfiguration | undefined = this.getInputConfiguration(`${id}`)
```

MadsThy (approved):

tsconfig.tsbuildinfo
MadsThy, line 1: I think this file goes on the gitignore list 😄

MadsThy (approved):

src/business-logic/interfaces/panel.ts
```diff
@@ -1,11 +1,77 @@
...
+
+const MODIFIER_DELIMITER: string = ';'
+
+export abstract class Panel {
+  public abstract initialize(): void
+  public abstract disconnect(): void
+  protected abstract assertValidPanelConfiguration(panelConfiguration: PanelConfiguration): void
+
+  protected activeModifiers: Set<PanelInputModifier> = new Set()
+  private modifierInputKeys: Set<string> = new Set()
```
KvelaGorrrrnio: You can consider using `ReadonlySet` instead :)

src/business-logic/interfaces/panel.ts
```diff
@@ -1,11 +1,77 @@
...
+
+  protected onCommandCallback?: (command: PanelCommand) => void
+
+  protected constructor(
+    protected readonly panelConfiguration: PanelConfiguration,
+    protected panelLayoutConfiguration: PanelLayoutConfiguration,
+    private readonly statusMessageService: StatusMessageService) {
```
KvelaGorrrrnio: It would be nice with the `) {` on a new line. I was confused why we had a statement as a parameter :P

src/model/enums/panel-enums.ts
```diff
@@ -22,3 +22,10 @@ export enum PanelCommandType {
   MODIFIER = 'MODIFIER',
   T_BAR = 'T_BAR',
 }
+
+export enum PanelInputModifier {
+  MODIFIER_ONE = 'MODIFIER_ONE',
+  MODIFIER_TWO = 'MODIFIER_TWO',
+  MODIFIER_THREE = 'MODIFIER_THREE',
+  MODIFIER_FOUR = 'MODIFIER_FOUR'
+}
```
KvelaGorrrrnio: These shouldn't be necessary. Having modifiers be  `string` types,  will enable user-defined modifiers.

src/business-logic/interfaces/panel.ts
```diff
@@ -1,11 +1,77 @@
...
-  updatePanelLayoutConfiguration(panelLayoutConfiguration: PanelLayoutConfiguration): void
-  registerOnCommand(onCommandCallback: (command: PanelCommand) => void): void
+import { InputConfiguration, PanelCommand } from '../../model/interfaces/input-configuration'
+import { PanelCommandType, PanelInputModifier } from '../../model/enums/panel-enums'
+import { StatusMessageService } from '../services/status-message-service'
+import { StatusMessage } from '../../model/entities/status-message'
+
+const MODIFIER_DELIMITER: string = ';'
+
+export abstract class Panel {
```
KvelaGorrrrnio, line 10: I would prefer if we try to limit the amount of inheritance. We can use composition by having a `PanelGroup`, where most of this can be placed. Then the modifier keys never has to reach the PanelManager and the only information reaching the PanelManager can be the PanelCommands.
Disclaimer: I haven't thought it entirely through, so it is something that can be brainstormed a bit.

src/business-logic/panel-integrations/skaarhoj-panel.ts
KvelaGorrrrnio: Or if we never convert it to a number: https://github.com/tv2/sofie-tv2-server/pull/44#discussion_r1825476089

KvelaGorrrrnio (commented):
I am not through the PR, but am stopping for the weekend :)

src/business-logic/panel-integrations/skaarhoj-panel.ts
LindvedKrvang: It has been turned into a string in the comment that Anders linked.

src/model/enums/panel-enums.ts
LindvedKrvang: True. That would also allow for more flexibility regarding modifiers. It's been updated

src/business-logic/interfaces/panel.ts
LindvedKrvang: Good catch. I completely agree :)

src/business-logic/interfaces/panel.ts
LindvedKrvang: Updated

src/business-logic/interfaces/panel.ts
LindvedKrvang: Looking into it. I agree that we should limit the amount of inheritance.

LindvedKrvang: I have introduced `PanelGroup` that contains the active modifiers for a group. 
I did not remove the inheritance as per our discussion.

RasmusAlbrektsen (approved):

src/business-logic/interfaces/panel.ts
```diff
@@ -1,11 +1,78 @@
...
+    }
+
+    const inputConfigurationKeys: string[] = Object.keys(this.panelLayoutConfiguration.inputConfigurations)
+    const inputKey: string | undefined = inputConfigurationKeys.find((key) => {
+      const separatedKey: string[] = key.split(MODIFIER_DELIMITER)
+      const doesKeyIncludeAllModifiers: boolean = Array.from(this.activeModifiers).every(activeModifier => separatedKey.includes(activeModifier))
+      const doesKeyIncludeInput: boolean = separatedKey.some(key => key === inputConfigurationKey)
+      const doesKeyHaveOtherModifiersThanActiveModifiers: boolean = separatedKey.length !== this.activeModifiers.size + 1
+
+      return doesKeyIncludeAllModifiers && doesKeyIncludeInput && !doesKeyHaveOtherModifiersThanActiveModifiers
```
KvelaGorrrrnio, line 66: Just a note: This computes everything before checking if some of these are false.
Using short-circuiting would be a nice addition to this. It make the method a bit longer.
This is also a good candidate to be moved to a separate method :)

src/business-logic/interfaces/panel.ts
```diff
@@ -1,11 +1,78 @@
...
+  public updateActiveModifiers(activeModifiers: ReadonlySet<string>): void {
+    this.activeModifiers = activeModifiers
+  }
+
+  public updatePanelLayoutConfiguration(panelLayoutConfiguration: PanelLayoutConfiguration): void {
+    this.panelLayoutConfiguration = panelLayoutConfiguration
+    this.updateModifierInputKeys()
+  }
+
+  private updateModifierInputKeys(): void {
```
KvelaGorrrrnio, line 46: Is there a difference between a _modifier_ and a _modifier input_ and _ modifier input keys_?

src/business-logic/panel-integrations/panel-group.ts
```diff
@@ -0,0 +1,66 @@
...
+    panel.registerOnCommand(panelCommand => this.handlePanelCommand(panelCommand))
+  }
+
+  private handlePanelCommand(panelCommand: PanelCommand): void {
+    if (panelCommand.type !== PanelCommandType.MODIFIER) {
+      this.onCommandCallback?.(panelCommand)
+      return
+    }
+
+    this.updateActiveModifiers(panelCommand.modifier)
```
KvelaGorrrrnio, line 26: I suggest that we for the modifier command specifies whether we are entering or leaving it, so we don't have to toggle. It works fine with one panel, but if we have two panels, which can enable/disable the same modifier, then it can give unexpected behavior.

src/business-logic/panel-integrations/skaarhoj-panel.ts
```diff
@@ -217,15 +203,27 @@ export class SkaarhojPanel implements Panel {
...
-        return skaarhojKeyEvent === inputConfiguration.triggersOn ? inputConfiguration.command : undefined
+
+        if (skaarhojKeyEvent !== inputConfiguration.triggersOn) {
+          return
+        }
+
+        if (inputConfiguration.command.type === PanelCommandType.MODIFIER) {
+          return {
+            ...inputConfiguration.command,
+            panelGroupId: this.panelConfiguration.panelGroupId
```
KvelaGorrrrnio, line 222: Is the `panelGroupId` necessary?

src/business-logic/panel-integrations/panel-group.ts
LindvedKrvang: There is going to be "latch" behaviour in https://tv2cms.atlassian.net/browse/SOF-2328 which is going to have something similar. So I would wait with this comment until that task

src/business-logic/panel-integrations/skaarhoj-panel.ts
LindvedKrvang: It appears not. I think it is a leftover from my initial implementation. It has been removed

src/business-logic/interfaces/panel.ts
LindvedKrvang: Yes.

`modifier` is whatever value that has been used as the actual modifier. This could be `MODIFIER_ONE`.

`modifierInput` is the configuration of the `InputConfiguration` that holds the modifier.

`modifierInputKeys` are the keys (or ids) of the actual inputs/buttons that is configured to be a ModifierInput. These are used so that we can deactivate a modifier again.
- Say we hit button 1, which is a modifier input. This input now activates the modifier "SOME_MODIFIER". If we don't save the key of the modifier, the next time we hit the same button we now try to search for an InputConfiguration for the `SOME_MODIFIER;1`, which doesn't exist.

Take the following object:
```
'46' : {
  "type": "BUTTON",
  "command": {
    "type": "MODIFIER",
    "modifier": "MODIFIER_ONE",
  },
  "triggersOn": "PRESSED"
}
```

The `'46'` is the `ModifierInputKey`. The object is the `ModifierInput`. The `MODIFIER_ONE` is the actaul modifier value.

src/business-logic/interfaces/panel.ts
LindvedKrvang: Refactored to "return early". Extracted it to a separate function

KvelaGorrrrnio (approved):

PR #47, opened 2024-11-01
Sof 2265/connect to atem directly

LindvedKrvang:
Note: Dependent on https://github.com/tv2/sofie-tv2-server/pull/46
Note: This PR also contains SOF-2266 and SOF-2267.

We now connect directly to Atem and updates the TBar position for FaderInputConfigurations for TBar.
Note: In this implementation the IP of the Atem is hardcoded. That will be address in my next PR. This PR is so the rest of the code can start being reviewed.


MadsThy (approved):

RasmusAlbrektsen (approved):

src/business-logic/panel-integrations/atem-video-mixer.ts
```diff
@@ -0,0 +1,88 @@
...
+      this.statusMessageService.sendStatusMessage(this.createConnectedStatusMessage())
+    })
+    this.atem.on('disconnected', () => this.reconnect())
+  }
+
+  private createConnectedStatusMessage(): StatusMessage {
+    return {
+      id: this.getStatusMessageId(),
+      title: 'Connected to Atem',
+      message: `Connected directly to Atem on ${ATEM_IP}. Ready to send T-Bar commands.`,
```
KvelaGorrrrnio: We can consider to specialize the title/message a bit more to not confuse the user. The title might be confusing for a user if this is good, but playout gateways connection to Atem is bad.
My best suggestion at to moment is to make this more about T-bar functionality, e.g.:
```typescript
      title: 'Atem is ready for T-bar transitions',
      description: `Connection established to Atem on ${ATEM_IP}. Ready to send T-bar transitions.`,
```

src/business-logic/panel-integrations/atem-video-mixer.ts
```diff
@@ -0,0 +1,88 @@
...
+    return {
+      id: this.getStatusMessageId(),
+      title: 'Reconnecting to Atem',
+      message: `Reconnecting to Atem on ${ATEM_IP}. Unable to send T-Bar command.`,
+      statusCode: StatusCode.UNKNOWN,
+      lastUpdatedTimestamp: Date.now()
+    }
+  }
+
+  public sendTBarCommand(tBarPosition: number): void {
```
KvelaGorrrrnio: There is a mixed responsibility here. Atem knows nothing about T-bars. What it exposes should be something like `setTransitionPosition` and then the caller is responsible to determine what value to use, whether that is the position of the T-bar, a fader or determined by a sequence of button clicks.

src/business-logic/panel-integrations/atem-video-mixer.ts
```diff
@@ -0,0 +1,88 @@
...
+
+      this.connect()
+    }, RECONNECTION_TIMEOUT_MS)
+  }
+
+  private createReconnectStatusMessage(): StatusMessage {
+    return {
+      id: this.getStatusMessageId(),
+      title: 'Reconnecting to Atem',
+      message: `Reconnecting to Atem on ${ATEM_IP}. Unable to send T-Bar command.`,
```
KvelaGorrrrnio: Same comment as for the connected status message. I don't have a good suggestion at the moment for making this differentiable from the playout Atem device status messages.

src/business-logic/panel-integrations/interfaces/video-mixer.ts
```diff
@@ -0,0 +1,3 @@
+export interface VideoMixer {
+  sendTBarCommand(tBarPosition: number): void // The tBarPosition should range from 0 to 1.000
```
KvelaGorrrrnio: Is it `1000` or `1` that is the upper limit?
If 1000, then why not use a normalized range from 0-1?

And just a repetition of my previous comment: The `VideoMixer` should not know anything about T-bars :)

src/business-logic/panel-integrations/skaarhoj-panel.ts
```diff
@@ -254,4 +266,30 @@ export class SkaarhojPanel extends Panel {
...
+
+    if (value > T_BAR_UPPER_BOUND) {
+      return T_BAR_UPPER_BOUND
+    }
+
+    if (value < T_BAR_LOWER_BOUND) {
+      return T_BAR_LOWER_BOUND
+    }
+    return value
+  }
```
KvelaGorrrrnio: The following suggestion is based on the assumption that the `value` that is given is the directed value always going from 0 to 1000 independent of the direction.
Suggestion: This method does 2 things. The first is to convert a directed value to a position and the second being setting upper and lower limits. The limits can also be set with `Math.min` and `Math.max`. E.g.:
```typescript
  private getTBarPosition(value: number): number {
    const tBarPosition: number = this.tBarDirection === TBarDirection.DOWN ? T_BAR_UPPER_BOUND - value : value
    return Math.min(T_BAR_UPPER_BOUND, Math.max(T_BAR_LOWER_BOUND, position))
  }
```

src/business-logic/panel-integrations/skaarhoj-panel.ts
```diff
@@ -254,4 +266,30 @@ export class SkaarhojPanel extends Panel {
...
+  private updateTBarDirection(value: number): boolean {
+    const upperBoundReached: boolean = value === T_BAR_UPPER_BOUND && this.tBarDirection === TBarDirection.UP
+    const lowerBoundReached: boolean = value === T_BAR_LOWER_BOUND && this.tBarDirection === TBarDirection.DOWN
+
+    if (upperBoundReached || lowerBoundReached) {
+      this.tBarDirection = upperBoundReached ? TBarDirection.DOWN : TBarDirection.UP
+      return true
+    }
+    return false
+  }
```
KvelaGorrrrnio: It looks like this works on the directed value (going from 0-1000), so shouldn't this always change on the T_BAR_UPPER_BOUND?
If so then this can be rewritten as the following to make the change meaningly observable from the returned result and remove side-effects:
```typescript
  private getTBarDirection(directedTBarValue: number): TBarDirection {
    if (directedTBarValue < T_BAR_UPPER_BOUND) {
      return this.tBarDirection
    }
    return this.tBarDirection === TBarDirection.DOWN ? TBarDirection.UP ? TBarDirection.DOWN
  }
```

Then it can be used as:
```typescript
const tBarDirection: TBarDirection = this.getTBarDirection(value)
inputConfiguration.command.shouldExecuteTake = tBarDirection !== this.tBarDirection
this.tBarDirection = tBarDirection

Comment (LindvedKrvang) [reply src/business-logic/panel-integrations/skaarhoj-panel.ts]: This method does not work on the directed value

File: src/business-logic/panel-integrations/interfaces/video-mixer.ts (hunk 4 above)
Comment (LindvedKrvang) [reply src/business-logic/panel-integrations/interfaces/video-mixer.ts]: It is `1000`. 
The simply reason for 1000 is because that's the range we get from the Skaarhoj Panel.

Updated the menions about T-bar.

File: src/business-logic/panel-integrations/atem-video-mixer.ts (hunk 3 above)
Comment (LindvedKrvang) [reply src/business-logic/panel-integrations/atem-video-mixer.ts]: Updated the title to be more T-bar specific

File: src/business-logic/panel-integrations/atem-video-mixer.ts (hunk 2 above)
Comment (LindvedKrvang) [reply src/business-logic/panel-integrations/atem-video-mixer.ts]: Technically, the transition position _is_ the position of the T-bar that is on the Atem. So it does know about T-bars ;) 
image

However, I agree with you. Updated to `setTransitionPosition` :)

File: src/business-logic/panel-integrations/atem-video-mixer.ts (hunk 1 above)
Comment (LindvedKrvang) [reply src/business-logic/panel-integrations/atem-video-mixer.ts]: Updated

File: src/business-logic/panel-integrations/skaarhoj-panel.ts (hunk 5 above)
Comment (LindvedKrvang) [reply src/business-logic/panel-integrations/skaarhoj-panel.ts]: Updated to try your method

File: src/business-logic/panel-integrations/skaarhoj-panel.ts (hunk 6 above)
Comment (LindvedKrvang) [reply src/business-logic/panel-integrations/skaarhoj-panel.ts]: Updated to try your method

File: src/business-logic/panel-integrations/atem-video-mixer.ts (hunk 2 above)
Comment (KvelaGorrrrnio) [reply src/business-logic/panel-integrations/atem-video-mixer.ts]: I stand corrected :P but, we don't do it for the virtual t-bar 😁

File: src/business-logic/panel-integrations/atem-video-mixer.ts [hunk 7]
```diff
@@ -0,0 +1,88 @@
...
+      }
+
+      this.connect()
+    }, RECONNECTION_TIMEOUT_MS)
+  }
+
+  private createReconnectStatusMessage(): StatusMessage {
+    return {
+      id: this.getStatusMessageId(),
+      title: 'Unable to sent T-bar transitions to Atem',
```
KvelaGorrrrnio: Shouldn't `sent` be `send`? :)

src/business-logic/panel-integrations/interfaces/video-mixer.ts
KvelaGorrrrnio: Could we update `1.000` to `1000` then. It can be a source of confusion, since if this was written as code it would be `1`.

src/business-logic/panel-integrations/skaarhoj-panel.ts
```diff
@@ -254,4 +257,25 @@ export class SkaarhojPanel extends Panel {
 
     return undefined
   }
+
+  private updateTBarCommandWithValues(command: TBarPanelCommand, value: number): PanelCommand {
+    const tBarValue: number = this.getTBarPosition(value)
+    const tBarDirection: TBarDirection = this.getTBarDirection(tBarValue)
```
KvelaGorrrrnio: There seem to be a naming issue here, which is most likely derived from my previous comments :D
We have a `value`, which we convert to a T-bar position. The position is then given to the method `getTBarDirection` that expects a `directedTBarValue` - which is counterintuitive.
Since the code works, then it points to that `getTBarPosition` actually calculates the directed T-bar value instead of the position and that the given value is the T-bar position.

An alternative name for the directedTBarValue could be `tBarTransitionProgress`.

So should this be something like?
```typescript
  private updateTBarCommandWithValues(command: TBarPanelCommand, tBarPosition: number): PanelCommand {
    const tBarTransitionProgress: number = this.getTBarTransitionProgress(tBarPosition)
    const tBarDirection: TBarDirection = this.getTBarDirection(tBarTransitionProgress)
```

src/model/interfaces/input-configuration.ts
```diff
@@ -32,12 +32,12 @@ export interface ActionPanelCommand extends BasePanelCommand {
 export interface ModifierPanelCommand extends BasePanelCommand {
   type: PanelCommandType.MODIFIER
   modifier: string
-  panelGroupId?: string
 }
 
 export interface TBarPanelCommand extends BasePanelCommand {
   type: PanelCommandType.T_BAR
-  value?: number
+  value?: number // The number range should go from 0 to 1.000
```
KvelaGorrrrnio: Same comment here for the range: `1.000` can easily be read as `1`. So should it be `1000`?

KvelaGorrrrnio (approved):
Looks good :)

src/business-logic/panel-integrations/skaarhoj-panel.ts

src/model/interfaces/input-configuration.ts
LindvedKrvang: Updated. Also at the other comment

src/business-logic/panel-integrations/atem-video-mixer.ts
LindvedKrvang: Updated

PR #54, opened 2024-11-04
Sof 2265/fetch video mixer configuration from alba server

LindvedKrvang:
Note: Dependent on https://github.com/tv2/sofie-tv2-server/pull/47 and https://github.com/tv2/sofie-server/pull/272

We now fetch the VideoMixerConfiguration from AlbaServer to find the hostname and port of the VideoMixer.
We also listen for updates to the VideoMixerConfiguration, so that we can update to the VideoMixer whenever it is reconfigured.

PR #55, opened 2024-11-04
Sof 2269/update rundown

LindvedKrvang:
Note: Dependent on https://github.com/tv2/sofie-tv2-server/pull/54

We now fetch the active RundownId when starting Tv2Server and uses it to send commands about the correct Rundown to AlbaServer.

We also listen for ACTIVATED, REHEARSE and DEACTIVATED events to keep the active RundownId up to date.

Also in this PR:
- Wrapped the `handleCommand` in the PanelManager in a try-catch so the system doesn't crash if we throw an error from in there.


src/presentation/fastify-server.ts
```diff
@@ -127,6 +130,15 @@ export class FastifyServer implements ProxyServer {
...
       this.deviceEmitter.emitVideoMixerConfigurationUpdated(parsedData.videoMixer)
     }
+
+    if (this.isRundownEvent(parsedData)) {
+      if (parsedData.type === RundownEventType.ACTIVATED || parsedData.type === RundownEventType.REHEARSE) {
+        this.rundownEmitter.emitActiveRundownId(parsedData.rundownId)
+      }
+      if (parsedData.type === RundownEventType.DEACTIVATED) {
+        this.rundownEmitter.emitActiveRundownId(undefined)
+      }
```
MadsThy:
```suggestion
      if (parsedData.type === RundownEventType.DEACTIVATED) {
        this.rundownEmitter.emitActiveRundownId(undefined)
        return
      }
      this.rundownEmitter.emitActiveRundownId(parsedData.rundownId)
```

Or perhaps this one :)

```suggestion
      parsedData.type === RundownEventType.DEACTIVATED) ? this.rundownEmitter.emitActiveRundownId(undefined) : this.rundownEmitter.emitActiveRundownId(parsedData.rundownId)
```

MadsThy (approved):

RasmusAlbrektsen (approved):

src/business-logic/interfaces/rundown-observer.ts
```diff
@@ -0,0 +1,3 @@
+export interface RundownObserver {
+  subscribeToRundownActiveRundownId(onActiveRundownIdChangedCallback: (rundownId: string | undefined) => void): void
```
KvelaGorrrrnio: Wouldn't `subscribeToActiveRundownId` suffice? :)

src/presentation/fastify-server.ts
```diff
@@ -127,6 +130,15 @@ export class FastifyServer implements ProxyServer {
     if (this.isEventVideoMixerConfigurationUpdatedEvent(parsedData)) {
       this.deviceEmitter.emitVideoMixerConfiguration(parsedData.videoMixer)
     }
+
+    if (this.isRundownEvent(parsedData)) {
```
KvelaGorrrrnio: You can avoid a layer of nesting by making this into a guard that returns.

src/presentation/fastify-server.ts
KvelaGorrrrnio: Since it is only the `type` that is checked then a switch could also be used :)

KvelaGorrrrnio (approved):

src/presentation/fastify-server.ts
LindvedKrvang: To use the switch we would need to make some more drastic changes to the overall method, else we would have a mixture of if-statements and a switch in the same method that all check on the type.

src/presentation/fastify-server.ts
LindvedKrvang: I think the idea was to encapsulate each "type" of event within its own scope. I have updated it to use the guard clause. 
I'm not entirely sastified with how the `fastify-server` handles different responsibilities, so it would be nice if we took a proper look at it at some point.

src/business-logic/interfaces/rundown-observer.ts
LindvedKrvang: You're absolutely right. This looks like a brainfart from my side. Good catch :) 

It's been updated

PR #56, opened 2024-11-04
SOF-2271 Send commands back to Skaarhoj panel

LindvedKrvang:
Note: Dependent on https://github.com/tv2/sofie-tv2-server/pull/55

Able to send "commands" back to the Skaarhoj panel.
For now we can "turn buttons on/off" (should the show light or not), set the color of a button and set text in the display fields.

Added `color: Color` and `label: InputLabel` to `BaseInputConfiguration`, so that it's possible to set the color and text of inputs.

"SkaarCommands" has been introduced. This is a way to abstract away the actual string that is sent to Skaarhoj (it could look like `HWCc#12=132`). Instead we get to work with nice classes (the earlier string is the same as `new SkaarhojColorCommand('12', Color.RED)`


src/business-logic/panel-integrations/skaarhoj-command.ts
```diff
@@ -0,0 +1,118 @@
...
+  BLUE = '139',
+  CYAN = '141',
+  GREEN = '196'
+}
+
+export abstract class SkaarhojCommand {
+  protected constructor(protected id: string) {}
+
+  public abstract toString(): string
+}
```
KvelaGorrrrnio, line 32: Just a thought: Couldn't this be an interface like:
```typescript
export interface SkaarhojCommand {
  readonly id: string
  toString(): string
}
```
And is the `id` necessary for all commands?

KvelaGorrrrnio (approved):

RasmusAlbrektsen (approved):

src/business-logic/panel-integrations/skaarhoj-command.ts
LindvedKrvang: The id is the identifer of the input the command is going to "modify". How would you go about identifying the input without some form of id?

For the interface thing: In the current implementation it currently doesn't matter that much whether it's an interface or abstract class.

src/business-logic/panel-integrations/skaarhoj-panel.ts
```diff
@@ -244,6 +260,7 @@ export class SkaarhojPanel extends Panel {
         return this.updateTBarCommandWithValues(inputConfiguration.command, value)
       }
     }
+    return
```
MadsThy, line 263: i don't think i see what this return does 😄

MadsThy (approved):

src/business-logic/panel-integrations/skaarhoj-panel.ts
LindvedKrvang: The return type of the method isn't `void`. When the return type isn't `void` the method has to have a return statement.

PR #58, opened 2024-11-11
SOF-2272 Update 'GUI' on Rundown activation/deactivation

LindvedKrvang:
Note: Dependent on https://github.com/tv2/sofie-tv2-server/pull/57

We now show the name of the active Rundown in the Panel. For now the display used for this has been hardcoded pr Panel.
When the Rundown is deactivated, we "clear" the Panel (not showing any colors etc and clicking inputs have no effect) and show a message saying "No active Rundown".
Upon Rundown activation we show all colors and text in the Panel again and clicking inputs works.


src/business-logic/services/rundown-http-service.ts
```diff
@@ -0,0 +1,19 @@
...
+import { RundownService } from '../interfaces/rundown-service'
+
+const RUNDOWN_ENDPOINT: string = '/rundowns'
+const BASIC_RUNDOWNS_ENDPOINT: string = `${RUNDOWN_ENDPOINT}/basic`
+
+export class RundownHttpService implements RundownService {
+  public constructor(private readonly httpService: HttpService) {}
+
+  public async getRundown(rundownId: string): Promise<Rundown> {
+    return await this.httpService.get(`${RUNDOWN_ENDPOINT}/${rundownId}`) as Rundown
```
MadsThy: I just had a thought :) How would you like to type your HTTP calls like this instead?

```suggestion
    return await this.httpService.get<Rundown>(`${RUNDOWN_ENDPOINT}/${rundownId}`)
```

I made this change to the HttpService to achieve that :)

```
get<T>(url: string): Promise<T>
```

MadsThy (approved):

src/business-logic/panel-integrations/panel.ts
```diff
@@ -12,11 +13,14 @@ export abstract class Panel {
...
   protected abstract assertValidPanelConfiguration(panelConfiguration: PanelConfiguration): void
   protected abstract clearPanelState(): void
-  protected abstract sendPanelState(): void
+  protected abstract sendActivePanelState(): void
+  protected abstract sendInactivePanelState(): void
 
   protected activeModifiers: ReadonlySet<string> = new Set()
   private modifierInputKeys: ReadonlySet<string> = new Set()
 
+  protected activeRundown: Rundown | undefined
```
KvelaGorrrrnio, line 22: Just a note our use of terms for rundown modes: Does this refer to a rundown that is ready for execution or a rundown in _active_ mode? :P

src/business-logic/panel-integrations/skaarhoj-panel.ts
```diff
@@ -321,4 +327,19 @@ export class SkaarhojPanel extends Panel {
 
     return commands
   }
+
+  protected override sendInactivePanelState(): void {
```
KvelaGorrrrnio: Why is this marked with `override`?

src/business-logic/services/panel-command-executor.ts
```diff
@@ -25,17 +24,17 @@ export class PanelCommandExecutor {
...
     this.logger = logger.tag(PanelCommandExecutor.name)
     this.rundownObserver.subscribeToActiveRundownId(rundownId => this.activeRundownId = rundownId)
-    this.fetchActiveRundown().catch(error => this.logger.data(error).error('Failed to fetch active Rundown'))
+    this.fetchActiveRundownId().catch(error => this.logger.data(error).error('Failed to fetch active Rundown'))
   }
 
-  private async fetchActiveRundown(): Promise<void> {
-    const rundowns: Rundown[] = await this.httpService.get(BASIC_RUNDOWN_ENDPOINT) as Rundown[]
-    const activeRundown: Rundown | undefined = rundowns.find(rundown => rundown.mode === RundownMode.ACTIVE || rundown.mode === RundownMode.REHEARSAL)
+  private async fetchActiveRundownId(): Promise<void> {
```
KvelaGorrrrnio: Suggestion: If the rundown is given along with command to the `executeActionCommand` method, then we won't have to maintain rundown state here as well.

src/business-logic/panel-integrations/panel.ts
```diff
@@ -28,6 +32,15 @@ export abstract class Panel {
...
   }
 
+  protected updatePanelState(): void {
+    this.clearPanelState()
+    if (!this.activeRundown) {
+      this.sendInactivePanelState()
+    } else {
+      this.sendActivePanelState()
+    }
+  }
```
KvelaGorrrrnio, line 42: Just a note: This could also be split up into a gather and act phases, where we first determine the state and then send the state. That way we can also keep the same level of detail throughout the  method and methods become more composable and testable.
```typescript
protected updatePanelState(): void {
  this.clearPanelState()
  const panelStateCommands: PanelStateCommand[] = this.getPanelStateCommands(this.activeRundown)
  this.sendCommands(panelStateCommands)
}
```

src/business-logic/services/rundown-http-service.ts
```diff
@@ -0,0 +1,19 @@
...
+  }
+
+  public async getActiveRundown(): Promise<Rundown | undefined> {
+    const rundowns: Rundown[] = await this.httpService.get(BASIC_RUNDOWNS_ENDPOINT) as Rundown[]
+    return rundowns.find(rundown => rundown.mode === RundownMode.ACTIVE || rundown.mode === RundownMode.REHEARSAL)
```
KvelaGorrrrnio, line 17: This seems to be a basic rundown that is returned, not a rundown.

src/business-logic/services/rundown-http-service.ts
KvelaGorrrrnio: The `await` is not necessary :)

RasmusAlbrektsen (approved):

src/business-logic/services/rundown-http-service.ts
LindvedKrvang: Currently, Tv2Server doesn't know about "basic" and non-basic Rundowns.
All it knows about a Rundown is this:
image

Do we want to also make the distinction in Tv2Server? It might be necessary at some point, but I had hoped to avoid it for as long as possible.

src/business-logic/panel-integrations/panel.ts
LindvedKrvang: It refers to the Rundown currently being "worked on" i.e. whatever Rundown at is active or in rehearsal.
I'm open for naming suggestions :)

src/business-logic/panel-integrations/skaarhoj-panel.ts
LindvedKrvang: Because it is an abstract method declared on the parent class `Panel`. 
The compiler, however, doesn't seem to complain if I remove the declaration which I find weird...
Personally, I would like to keep the `override` to show that it comes from the parent class.

LindvedKrvang: Or... I guess that that is what `protected` indicates. In that case I'm fine with removing the `override` :)

LindvedKrvang: It's been removed

src/business-logic/services/panel-command-executor.ts
LindvedKrvang: Updated it so it is no longer the responsibility of the PanelCommandExecutor

src/business-logic/services/rundown-http-service.ts
LindvedKrvang: Something disagrees with you 
image

src/business-logic/panel-integrations/panel.ts
LindvedKrvang: This requires that we have a common `command` abstraction that we enforce for all our Panel integrations. I'm not sure that commands to different Panels has much in commen. For example, Skaarhoj commands needs to be converted into strings. Is that case for all other Panels as well?
We could have an empty abstraction, but then we would need to cast to the specific Panel command within the xPanel class.
All of the above might be irrelevant, but it is atleast the reason for why I kept the commands within the subclass for now :) 
I do however like spiltting the gathering of commands and sending into different methos :)

src/business-logic/services/rundown-http-service.ts
LindvedKrvang: It's been removed

src/business-logic/services/rundown-http-service.ts
KvelaGorrrrnio: The issue I am seeing here is that we cast a BasicRundown from alba-server to a Rundown, which is a reduced defintion here. But when we need to increase the model of the rundown, will we then remember to change assumption made here?

src/business-logic/services/panel-manager-implementation.ts
```diff
@@ -87,14 +98,16 @@ export class PanelManagerImplementation implements PanelManager {
   }
 
   private handleCommand(command: PanelCommand): void {
+    // Note: Since we assert the activeRundown here, it is safe for us to use the "this.activeRundown!" later in this method.
+    this.assertActiveRundown()
```
KvelaGorrrrnio, line 102: It is a bit more verbose, but one could change `assertActiveRundown` to take an optional rundown, then the `!` type assertion and comment won't be necessary:

```typescript
  private assertActiveRundown(rundown: Rundown | undefined): asserts rundown is Rundown {
    if (!rundown || !rundown.mode) {
      throw new NoActiveRundownException('Unable to execute command since there is no active Rundown')
    }
  }
```

src/business-logic/panel-integrations/panel.ts
KvelaGorrrrnio: The comment was mostly to create awareness that we are spreading this terminology to more projects, without the definition being clear. I think we should have a quick chat in the team about what to call rundown states and based on the outcome create a task for updating the lingo across the code bases.

src/business-logic/panel-integrations/panel.ts
KvelaGorrrrnio: Just a thought. I don't think it will be an issue. E.g. it could be that all panel commands should be serializable to a string. I think that would be generic enough for panels we will support. And the interface could easily be extended to serialize to a Buffer as well. Those two should cover the majority of protocols.

KvelaGorrrrnio (approved):

PR #59, opened 2024-11-11
SOF-2328 Implement Modifier latch behaviour

LindvedKrvang:
Note: Dependent on https://github.com/tv2/sofie-tv2-server/pull/58

Modifier keys now 'latch' when they are briefly pressed. If they are held down, they are active only for as long as they are held down.

Also in this PR:
- Added missing attribute `panelGroupId` to `PanelConfigurationDto`.
- If a modifier is active, the button is brighter to identicate it is active. 
- `InputConfiguration.triggersOn` is now an array of `KeyEvent` so that buttons can react on both PRESSED and RELEASED (which is needed for modifiers).


MadsThy (approved):

src/business-logic/panel-integrations/panel-group.ts
```diff
@@ -15,29 +17,50 @@ export class PanelGroup {
...
+  private updateModifierForKeyPressed(modifier: string): void {
     if (!this.activeModifiers.has(modifier)) {
-      this.activeModifiers.add(modifier)
-    } else {
-      this.activeModifiers.delete(modifier)
+      this.activeModifiers.set(modifier, Date.now())
+      return
+    }
+    this.activeModifiers.delete(modifier)
+  }
```
KvelaGorrrrnio: Is the `delete` used for latch release? An alternative could be to add a bit more state to the active modifiers. E.g.
```typescript
interface ModifierState {
  readonly pressedEpochTimestamp: number
  readonly releasedEpochTimestamp?: number // `isLatched: boolean` could also be used instead.
  readonly owner: string // The panel id, to determine if there are some presses/releases we should ignore in the case where the same modifier is pressed/released on multiple panels.
}
```
By having this being a toggle, it can give some funky behavior if multiple panels has a modifier key for the same modifier layer.

src/model/interfaces/input-configuration.ts
```diff
@@ -94,7 +94,7 @@ const ZOD_BASE_INPUT_CONFIGURATION_SCHEMA: ZodSchema<BaseInputConfiguration> = z
 
 const ZOD_BUTTON_CONFIGURATION_SCHEMA: ZodSchema<ButtonConfiguration> = z.intersection(ZOD_BASE_INPUT_CONFIGURATION_SCHEMA, z.object({
   type: z.literal(InputType.BUTTON),
-  triggersOn: z.nativeEnum(KeyEvent),
+  triggersOn: z.array(z.nativeEnum(KeyEvent)),
```
KvelaGorrrrnio: This creates a schema for a tuple with one value (`[KeyEvent]`). To accept an array of key events (`KeyEvent[]`), it should be `z.nativeEnum(KeyEvent).array()`.

KvelaGorrrrnio (commented):
Looks good :)

RasmusAlbrektsen (approved):

src/business-logic/panel-integrations/panel-group.ts
LindvedKrvang: It is.

Having multiple modifier buttons for the same modifier layer, but on different panels (even on the same panel) seems like an extreme edgecase, where someone has made a too convoluted configuration and can easily be fixed by making a more sensible configuration. (Also, when will a PanelGroup be operated by more than one person? It seems like a workflow doomed to fail if one person can suddenly change the input of what another person is doing without warning).
I would call YAGNI on it :)

src/model/interfaces/input-configuration.ts
LindvedKrvang: Was not aware. Good catch. It has been updated :)

src/business-logic/panel-integrations/panel-group.ts
KvelaGorrrrnio: I agree that it is an edge case, but I am also thinking of the maintainability of this. The placement is a bit counter intuitive. The release of a latched key happens when the key is pressed. I would assume that it would happen on release :)
If kept as is, I would say that this is one of the few cases where a comment describing why is appropriate. E.g.
```typescript
this.activeModifiers.delete(modifier) // The modifier is deleted to release a latched modifier.
```

src/model/interfaces/input-configuration.ts
```diff
@@ -18,7 +18,7 @@ export interface InputLabel {
 
 export interface ButtonConfiguration extends BaseInputConfiguration {
   type: InputType.BUTTON
-  triggersOn: KeyEvent
+  triggersOn: KeyEvent[]
```
KvelaGorrrrnio, line 21: Just a FP suggestion: The attribute and array can be made readonly. Same goes for the type attribute :)
```typescript
export interface ButtonConfiguration extends BaseInputConfiguration {
  readonly type: InputType.BUTTON
  readonly triggersOn: readonly KeyEvent[]
}
```

KvelaGorrrrnio (approved):

src/model/interfaces/input-configuration.ts
LindvedKrvang: FP? Functional programming?

src/business-logic/panel-integrations/panel-group.ts
LindvedKrvang: I like your point about releasing the latch on the RELEASE event. 
Turns out, all that was missing from having the latch be removed from the RELEASE event, was to not remove on the PRESSED event :) 

It's been updated

KvelaGorrrrnio (approved):

PR #75, opened 2025-01-17
Sof 2429/fixed server crash on deactivated rundown

vushu:
I added a guard for not handling commands when rundown is deactivated, instead of throwing i warn.


vushu:
whooops made PR to master .

PR #76, opened 2025-01-17
Added guard for handling commands only for when the rundown is active.

vushu:
I added a guard for handling commands only when the rundown is active, instead of throwing I log with a warning.


src/business-logic/services/panel-manager-implementation.ts
```diff
@@ -102,8 +101,11 @@ export class PanelManagerImplementation implements PanelManager {
   }
 
   private handleCommand(command: PanelCommand): void {
-    // Note: Since we assert the activeRundown here, it is safe for us to use the "this.activeRundown!" later in this method.
-    this.assertActiveRundown()
+    if (!this.activeRundown) {
```
LindvedKrvang, line 104: Since you are checking `!this.activeRundown` here, you can now remove the `!` from `this.activeRundown!` on line 112 and 116.
You removed the comment and the code that allowed us to use the `!` safely on those lines, so you should also remove the `!` as it is no longer needed :)

LindvedKrvang (approved):
Nice and simple :)

My approve assumes that you fix the one comment I added.

RasmusAlbrektsen (approved):

src/business-logic/services/panel-manager-implementation.ts
vushu: I have done as suggested :)

LindvedKrvang (approved):

PR #78, opened 2025-01-21
added reconnect on error

vushu:
It is very important to listen to error else the server will crash since it doesn't get caught!


LindvedKrvang (commented):
Testing this I was unable to get any information about TV2Server losing connection to alba-server. There isn't even anything in the logs... (I tried changing so it logged an error instead of a warning.)

The User would need to know that we have lost connection (and regainend it again). That could be a separate task, but there should _at least_ be something in the logs about the connection has been lost.

But, the TV2Server didn't crash when I stopped alba-server, and it reconnected to it when I started it again, so that functionality is working :)

vushu:
@LindvedKrvang can you point me to how to see the logs :)

vushu:
I made some changes which should hopefully display it now.

src/presentation/fastify-server.ts
```diff
@@ -107,21 +110,26 @@ export class FastifyServer implements ProxyServer {
...
     this.albaWebsocket?.close()
     this.albaWebsocket = new WebSocket(proxyConfiguration.websocketUrl)
-    this.albaWebsocket.addEventListener('open', () => this.logger.debug('Connected to AlbaServer'))
+    this.albaWebsocket.addEventListener('open', () => {
+      this.logger.info('Successfully connected to AlbaServer')
+    })
 
     this.albaWebsocket.addEventListener('close', () => {
-      this.logger.debug('Disconnected from Alba Server')
+      this.logger.info(`Failed to establish connection with AlbaServer, trying to reconnect in ${RECONNECT_DELAY_IN_MS / 1000} seconds ...`)
```
RasmusAlbrektsen: I'd prefer if we used `Alba Server` rather than `AlbaServer`.

src/presentation/fastify-server.ts
```diff
@@ -107,21 +110,26 @@ export class FastifyServer implements ProxyServer {
...
       setTimeout(() => this.connectToAlbaServer(proxyConfiguration), RECONNECT_DELAY_IN_MS)
     })
 
+    this.albaWebsocket.addEventListener('error', (err) => {
+      // Checking if err has any information.
+      if (Object.keys(err).length !== 0) {
```
RasmusAlbrektsen: I would change `err` to `error` and remove the comment on line 123, i don't think it's necessary. 
We usually try not to abbreviate, i understand this is not unusual for an error but keeping it as a rule of thumb i feel we avoid any possible confusion that could arise in other scenarios. :-)

RasmusAlbrektsen (approved):
Looks good, worked for me as well :-)

src/presentation/fastify-server.ts
vushu: I have added the requested changes.

src/presentation/fastify-server.ts
vushu: I have added the requested changes :)

RasmusAlbrektsen (approved):

src/presentation/fastify-server.ts
```diff
@@ -107,21 +110,25 @@ export class FastifyServer implements ProxyServer {
...
-      this.logger.debug('Disconnected from Alba Server')
+      this.logger.info(`Failed to establish connection with Alba Server, trying to reconnect in ${RECONNECT_DELAY_IN_MS / 1000} seconds ...`)
       setTimeout(() => this.connectToAlbaServer(proxyConfiguration), RECONNECT_DELAY_IN_MS)
     })
 
+    this.albaWebsocket.addEventListener('error', (error) => {
+      if (Object.keys(error).length !== 0) {
+        this.logger.data(error).error('An error occured.')
+      }
+    })
```
LindvedKrvang: Would be nice if you would include the type of `error`. It's `ErrorEvent`.

With the type specified then instead of the `Object.keys`, you could something like:
image

LindvedKrvang (approved):
Looks good.

Now we just need to notify the frontends when we lose and regain connection. But as mentioned before, I think that could be a separate task.

src/presentation/fastify-server.ts
vushu: I have done as suggested.

PR #80, opened 2025-01-29
Broadcast alba server connection status to clients

vushu:
I have added that we now broadcast the connection status of Alba server to the clients.
We notify on success and on fail to the clients, using the interface `StatusMessageEvent`

To test:

- Start Alba Server up
- Start TV 2 Server up 
- Start WebClient up

then shutdown Alba Server
The ui should get notified and viceversa.

I have manually tested that this works for both cases,


src/presentation/fastify-server.ts
```diff
@@ -112,10 +113,15 @@ export class FastifyServer implements ProxyServer {
     this.albaWebsocket = new WebSocket(proxyConfiguration.websocketUrl)
     this.albaWebsocket.addEventListener('open', () => {
       this.logger.info('Successfully connected to Alba Server')
+      const statusMessage: StatusMessageEvent = this.buildStatusMessageEvent({ statusCode: StatusCode.GOOD, message: 'We successfully established connection!', id: 'connected_to_alba_server', title: 'Connected to Alba Server.', lastUpdatedTimestamp: Date.now() })
```
RasmusAlbrektsen: I like the team spirit but i would suggest keeping the message to:
`Connection successfully established to server.`

In the title i would refrain from using `Alba Server` and just use `server`. At least in my opinion we don't need to inform in the user facing notifications about which specific server is down in the chain. But i would like opinions here? :-) @LindvedKrvang

src/presentation/fastify-server.ts
```diff
@@ -112,10 +113,15 @@ export class FastifyServer implements ProxyServer {
...
+      this.broadcastStatusMessageEvent(statusMessage)
     })
 
     this.albaWebsocket.addEventListener('close', () => {
-      this.logger.info(`Failed to establish connection with Alba Server, trying to reconnect in ${RECONNECT_DELAY_IN_MS / 1000} seconds ...`)
+      let retryMessage = `Retrying in ${RECONNECT_DELAY_IN_MS / 1000} seconds ...`
+      this.logger.warn(`Failed to establish connection with Alba Server. ${retryMessage}`)
+      const statusMessage: StatusMessageEvent = this.buildStatusMessageEvent({ statusCode: StatusCode.BAD, message: retryMessage, id: 'lost_connection_to_alba_server', title: 'Lost connection to Alba Server.', lastUpdatedTimestamp: Date.now() })
```
RasmusAlbrektsen: Same comment with `Alba Server` and just keeping it to `server`. But depends on what we decide in the previous comment of course!

RasmusAlbrektsen (approved):
Good stuff! Works for me too :-)

src/presentation/fastify-server.ts
vushu: @RasmusAlbrektsen I have no problem in dialing down the excitement level :), but about `server` vs `Alba server`,
The reason is because when we loose connection to the tv 2 server the webclient notifies with lost connection to backend, which doesn't really tell us much. I just added alba server since, then we can differentiate between connection status of tv2 server and the connection status of alba server. 

I believe there is a pro and con in this
if we should make these messages clear then we know which chain has been broken, easier for incident reporting.

We already specifically notify the webclient on inews gateway's connection status.
So we have message mismatch of specific and unspecific notifications.
Either we keep it specific or keep it unspecific.

RasmusAlbrektsen: I would argue that you already tell us this info through the logger, but i see the argument of a user being able to write it in the incident report more specifically.

LindvedKrvang: > We already specifically notify the webclient on inews gateway's connection status.
> So we have message mismatch of specific and unspecific notifications.
> Either we keep it specific or keep it unspecific.

There is a fundemental difference here. The iNews-gateway is not a requirement for the backend to work. It's not even an requirement for Alba to produce a show. It's just an auxillary (or device or whathever you wanna call it) where we get our Ingest data. This could, in theory, come from anywhere (SuperFly has made a gateway that gets the data from a Google Docs document).

In other words: The iNews-gateway is something that is _outside_ of the backend, hence when we lose connection to it we specify it.

Tv2Server and AlbaServer are integral parts of Alba and if one of them goes down or we lose connection to them, then Alba ceases to work. Hence they become the backend.

But to keep things simple. Who is this message for? It's for the User. The User has no idea of that there exists a Tv2Server and an AlbaServer. So specifying these servers is just going to confuse them. On the other hand, they do know that in order to get their Rundown from iNews, they need the iNews gateway.

So if a User gets an error saying "Lost connection to iNews", then they know that they can keep producing the show, but they won't be able to receive new data/updates from iNews.
If the User receives a "Lost connection to backend", then they know something bad has happened, and that they shouldn't expect to be able to produce the show.

src/presentation/fastify-server.ts
LindvedKrvang: I vote that we just show "the same message" as when the frontend lose its own connection.
image

The User don't get much value from knowing which specific server went down, because they most likely don't even know they exist.

Offcourse, keep the server specifications in the log statements for our use :)

src/presentation/fastify-server.ts
LindvedKrvang: I vote that the message could just be the same as the frontends message when it gains its own connection back 
image

Offcourse keep the specific information in the logs :)

src/presentation/fastify-server.ts
```diff
@@ -112,10 +113,15 @@ export class FastifyServer implements ProxyServer {
     this.albaWebsocket = new WebSocket(proxyConfiguration.websocketUrl)
     this.albaWebsocket.addEventListener('open', () => {
       this.logger.info('Successfully connected to Alba Server')
+      const statusMessage: StatusMessageEvent = this.buildStatusMessageEvent({ statusCode: StatusCode.GOOD, message: 'We successfully established connection!', id: 'connected_to_alba_server', title: 'Connected to Alba Server.', lastUpdatedTimestamp: Date.now() })
+      this.broadcastStatusMessageEvent(statusMessage)
     })
 
     this.albaWebsocket.addEventListener('close', () => {
-      this.logger.info(`Failed to establish connection with Alba Server, trying to reconnect in ${RECONNECT_DELAY_IN_MS / 1000} seconds ...`)
+      let retryMessage = `Retrying in ${RECONNECT_DELAY_IN_MS / 1000} seconds ...`
```
LindvedKrvang: Minor thing: Use `const` over `let` when the variable is never changed after initialization.

Also, would be nice if you specified the type of the variable. This specific case is a case where you could leave it out because it's quite evident that the variable is a string, but (in my personal opinion) it's a good idea to just always write the type :) 

Minor personal opinion - take it or leave it: I think I would exclude specifying how long it is until we try and reconnect again and simply say something like "Attempting to reconnect..."

src/presentation/fastify-server.ts
```diff
@@ -112,10 +113,15 @@ export class FastifyServer implements ProxyServer {
...
+      this.logger.warn(`Failed to establish connection with Alba Server. ${retryMessage}`)
```
LindvedKrvang: Wouldn't this be an error? If we don't have a connection to AlbaServer, then Alba as a whole stops working.

Isn't warning used for stuff where you could say: "Everything is still working, but be aware of this thing over here which isn't great and could lead to some bad stuff in the future"?

So a warning would be: "We connected to AlbaServer, but it took us 10 seconds which is quite bad and could indicate some issues with AlbaServer".

And an error would be: "We couldn't connect to AlbaServer"

src/presentation/fastify-server.ts
vushu: I have done the change to be `Reconnected to backend.`

src/presentation/fastify-server.ts
vushu: I made the change to being the one described.

src/presentation/fastify-server.ts
```diff
@@ -112,10 +114,25 @@ export class FastifyServer implements ProxyServer {
     this.albaWebsocket = new WebSocket(proxyConfiguration.websocketUrl)
     this.albaWebsocket.addEventListener('open', () => {
       this.logger.info('Successfully connected to Alba Server')
+      const statusMessage: StatusMessageEvent = this.buildStatusMessageEvent({
+        statusCode: StatusCode.GOOD,
+        message: '', id: ALBA_CONNECTION_STATUS_MESSAGE_EVENT_ID,
```
LindvedKrvang: Please either put all attributes on one line, or new-line every single attribute.

Also: It would be nice if `id` was the first attribute :)

src/presentation/fastify-server.ts
```diff
@@ -112,10 +114,25 @@ export class FastifyServer implements ProxyServer {
...
 
     this.albaWebsocket.addEventListener('close', () => {
-      this.logger.info(`Failed to establish connection with Alba Server, trying to reconnect in ${RECONNECT_DELAY_IN_MS / 1000} seconds ...`)
+      const retryMessage = `Retrying in ${RECONNECT_DELAY_IN_MS / 1000} seconds ...`
+      this.logger.error(`Failed to establish connection with Alba Server. ${retryMessage}`)
+      const statusMessage: StatusMessageEvent = this.buildStatusMessageEvent({
+        statusCode: StatusCode.BAD,
+        message: retryMessage,
+        id: 'alba_server_connection',
+        title: 'Attempting to reconnect...',
```
LindvedKrvang: Would be nice if it said `Attempting to reconnect to backend...` so that the user has an idea of what it is connection to :)

LindvedKrvang (approved):

PR #93, opened 2025-02-17
Added disable for panel configuration

vushu:
Since it disconnect after you update the panel as the Jira task explained.
I went with the simple implementation and it worked very well.
 implementation:
```typescript
private connectToPanel(panelConfiguration: PanelConfiguration): void {
  if (panelConfiguration.isDisabled) {
    return
  }
  // Code to connect to a Panel
}
```
I was poking around with postman and didn't hit any issues.


src/presentation/dtos/panel-configuration-dto.ts
```diff
@@ -16,5 +17,6 @@ export class PanelConfigurationDto {
     this.hostname = panelConfiguration.hostname
     this.panelGroupId = panelConfiguration.panelGroupId
     this.panelLayoutConfigurationId = panelConfiguration.panelLayoutConfigurationId
+    this.isDisabled = panelConfiguration.isDisabled ?? false
```
LindvedKrvang: The `?? false` is redundant. `isDisabled` isn't optional on PanelConfiguration, so the value will always be either `true` or `false`.

LindvedKrvang (approved):

RasmusAlbrektsen (approved):

src/presentation/dtos/panel-configuration-dto.ts
vushu: I have added a migration script for adding this field.

PR #107, opened 2025-03-21
SOF-2530 Add AuditLog

LindvedKrvang:
TV2 servers equivalent of https://github.com/tv2/alba-server/pull/331


RasmusAlbrektsen (approved):

src/presentation/decorators/audit-log-decorator.ts
```diff
@@ -0,0 +1,37 @@
...
+
+interface Request {
+  body?: unknown
+  params?: unknown
+}
+
+export function AuditLog(): MethodDecorator {
+  const logger: Logger = LoggerFacade.createLogger()
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
vushu, line 12: Could you be kind and briefly explain whats happening here :), I understand that you set the type to `any` since you can decorate this on any method?

vushu (approved):

src/presentation/decorators/audit-log-decorator.ts
LindvedKrvang: We have a linter-rule saying that we are not allowed to use the type `any`. This line disables that linter-rule for this specifc place so we can compile.

The reason for using the `any` (which we should strive to avoid using) is because of how `MethodDecorator` works. The `TypedPropertyDescriptor` expects a generic, which we can really provide here. But because we can't provide the generic we can't build our project. No matter what type we provide the actual value `could be a descendent of T`. 
The only way I have gotten the compiler to build is by using `any`...
I have tried using `unknown`, but I get the same compiler error when ubilding.

PR #111, opened 2025-04-03
Display active rundown name only for the MKT1A panel.

vushu:


src/business-logic/panel-integrations/skaarhoj-panel.ts
```diff
@@ -313,7 +313,7 @@ export class SkaarhojPanel extends Panel {
       return inputConfiguration ? this.mapInputConfigurationToSkaarhojCommands(inputId, inputConfiguration) : []
     })
 
-    if (this.activeRundown) {
+    if (this.activeRundown && this.panelConfiguration.model === SkaarhojModel.MKT1A) {
```
LindvedKrvang: The functionality is to not show in on the MK48 panel. If we had a third panel this change would also make it not shown on that panel.

This means that we should check that we are not a MK48 panel

LindvedKrvang (approved):
The approve assumes that the if-statement is "reversed"

LindvedKrvang (approved):

RasmusAlbrektsen (approved):

PR #116, opened 2025-04-25
Max deviation guard

vushu:
Draft possible fix for skaarhoj jumping values.


LindvedKrvang (commented):
It looks good.
However, this will break "fast takes" as the values might jump with over 500 from between each input from the Skaarhoj Panel. In the attached video, on the last take Skaarhoj only sends input three times to cover the jump from 0 to 1.000.

It seems the bug appears because Skaarhoj itself sends a value going from one extremity to another.

I will suggest an approach where if we are within, lets say 10% of one extremity and we suddenly jump to be within 10% of the another extremity, then we should consider it false input.


https://github.com/user-attachments/assets/2164a854-41a7-4ab1-9c63-b6bc8b8657a1

vushu:
I have tested this POC, with the panel that has this issue, I had no luck in fixing it so yeah back to the drawing board.

PR #117, opened 2025-04-29
SOF-2567 - Reconnect Skaarhoj panels after failed startup

RasmusAlbrektsen:
If the TV2 server started up prior to the Alba Server being up and and running, we would failed our initial starting process and never retry it, only re-attempt the direct connection to Alba Server. 

This PR allows us to retry with 5 second intervals so long as we fails the startup.

I put the retry around the entire `startAlbaTv2Server()` method to encompass all cases we might fail, but technically we could solve this specific bug task by just doing it with `startSystemServices()`.


vushu (approved):

LindvedKrvang (commented):
@RasmusAlbrektsen Have you checked what happens when the database connection and proxy server is started the second time? I.e. do we clear those connections when we make new ones, or do we now simply have two proxy servers (potentially more) running a the same time? (And database connections)

This solution works because right now `startSystemServices` only starts the PanelManager.
I am a bit concerned that we are restarting everything just because one service fails. 

Also, it seems a bit "aggresive/invasive" to restart the entire app because one service couldn't connect to a "third party peripheral" - A peripheral that might not even be on the network.

Couldn't the retry logic be in the `PanelManager.initialize` where we already have a try-catch?
image

RasmusAlbrektsen:
> @RasmusAlbrektsen Have you checked what happens when the database connection and proxy server is started the second time? I.e. do we clear those connections when we make new ones, or do we now simply have two proxy servers (potentially more) running a the same time? (And database connections)
> 
> This solution works because right now `startSystemServices` only starts the PanelManager. I am a bit concerned that we are restarting everything just because one service fails.
> 
> Also, it seems a bit "aggresive/invasive" to restart the entire app because one service couldn't connect to a "third party peripheral" - A peripheral that might not even be on the network.
> 
> Couldn't the retry logic be in the `PanelManager.initialize` where we already have a try-catch? image

Good points, i will test with a more isolated approach and see if i can cover our needs that way :-)

LindvedKrvang (approved):

PR #122, opened 2025-05-12
SOF-2576 Skaarhoj panels light up when their configured Action is invoked

LindvedKrvang:
Note: Dependent on https://github.com/tv2/alba-tv2-server/pull/121

Skaarhoj panels now ligt up its button if the configuration of the button has an Action that matches one of the "invoked" actions.


vushu (approved):
LGTM

RasmusAlbrektsen (approved):

PR #123, opened 2025-05-12
Sof 2572/made t bar value spike detection for faulty skaarhoj panel

vushu:
I have made a simple detection for when skaarhoj t-bar makes spikes in the value.
This way we know that there are some wear and tear on the t-bar.
The detection looks at how big a jump in value is so if the jump is greater than 950 then we will throw an exception.

I have made some tests, where I fake a skaarhoj panel, so that simulation of the t-bar spike values are possible.
Not sure if that is the right approach, feedback is welcome.

Also this detection isn't bullet proof, since the t-bar values is quite wobbly, and may jump not only from 
0 - 1000, 1000 - 0, which this detection will catch.

At times the T-bar wobbles from 
500 - 1000, 
930 - 0.

In these cases we won't be able to detect.

Argument for this PR is we still detect problem and log, so we know that it's a of a faulty panel and thereby we can notify Control Rooms for assistance.


src/business-logic/panel-integrations/skaarhoj-panel.ts
```diff
@@ -288,6 +298,16 @@ export class SkaarhojPanel extends Panel {
     return command
   }
 
+  private detectTbarSpike(value: number): void {
+    if (this.lastTbarValue !== undefined) {
+      const deviation = Math.abs(value - this.lastTbarValue)
+      if (deviation >= MAX_T_BAR_DEVIATION) {
+        throw new TBarValueSpikeException(`T-bar made a big spike of ${deviation}, this will cause issues.`)
```
RasmusAlbrektsen: Could it perhaps be:
```
throw new TBarValueSpikeException(`The T-bar made an unusual deviation in values. 
Deviation:  ${deviation}.`)
```

RasmusAlbrektsen (approved):

LindvedKrvang (approved):

src/business-logic/panel-integrations/test/skaarhoj-panel.spec.ts
```diff
@@ -1,12 +1,19 @@
...
 import { PanelConfiguration } from '../../../model/interfaces/panel-configuration'
 import { EntityTestFactory } from '../../../model/test/entity-test-factory'
-import { PanelModel, PanelType, SkaarhojModel } from '../../../model/enums/panel-enums'
+import { PanelCommandType, PanelModel, PanelType, SkaarhojModel } from '../../../model/enums/panel-enums'
 import { UnsupportedOperationException } from '../../../model/exceptions/unsupported-operation-exception'
 import { Logger } from '../../../logger/logger'
 import { instance, mock } from '@typestrong/ts-mockito'
 import { StatusMessageService } from '../../services/status-message-service'
 import { ColorConverter } from '../../interfaces/color-converter'
+import {  TBarPanelCommand, PanelCommand } from '../../../model/interfaces/input-configuration'
```
RasmusAlbrektsen: Just a few extra spaces in front of `TBarPanelCommand`

PR #124, opened 2025-05-14
Sof 2581/misc lights in skaarhoj

LindvedKrvang:
Note: Dependent on https://github.com/tv2/alba-server/pull/348

InvokedActions now also include inputs to be inserted and recall last splitscreen/remote.

If the `OutputChannel` is `UNKNOWN` we then check in both "output channels".


vushu (approved):
LGTM

RasmusAlbrektsen (approved):

PR #141, opened 2025-09-17
Ag 676 implement cal ver workflows in alba tv2 server

rana-tv2:


package.json
```diff
@@ -9,16 +9,15 @@
     "build": "yarn clean && tsc --build",
     "build-watch": "tsc --build --watch",
     "watch": "nodemon",
-    "test": "vitest run",
+    "test": "vitest run --coverage",
```
KvelaGorrrrnio: Can we move this flag to the vitest config? :D
Having all test configuration in one place makes it easier to see how it is configured and easier to port to other repositories. It is done by setting `test.coverage.enabled` to `true` in `vitest.config.mts`.

KvelaGorrrrnio (approved):
Nice work :D

package.json
rana-tv2: removed --coverage flag and add it in vitest.config.mts

KvelaGorrrrnio (approved):

